PR #15, opened 2015-09-04
New: typeahead feature

buchslava:


components/typeahead/typeahead.ts
```diff
@@ -0,0 +1,447 @@
...
+    </li>
+  </ul>
+  `
+};
+
+export class TypeaheadOptions {
+  public placement:string;
+  public animation:boolean;
+
+  constructor(options:Object) {
```
valorkin: `:TypeaheadOptions`

components/typeahead/typeahead.ts
```diff
@@ -0,0 +1,447 @@
...
+    Object.assign(this, options);
+  }
+}
+
+export class Match {
```
valorkin: @buchslava to small? may be not sense?

valorkin:
Closes #13

PR #17, opened 2015-09-15
Declare lifecycle events using implements keywoard

rkintzi:
Adaptation to the newest changes in angular-2.0.0-alpha.37:
https://github.com/angular/angular/blob/master/CHANGELOG.md#breaking-changes

Tested only with `alert` directive.


valorkin:
@rkintzi awesome, I will check other components

valorkin:
@rkintzi when updating ng2 version, zone.js and rtts_assert should be updated too (I fill fix)

PR #26, opened 2015-10-26
fix(pagination): updating Pagination.totalPages

ryzy:
When updating `Pagination.totalPages` after total amount of items
changed it goes to the last page (by calling
`this.selectPage(totalPages)`).

Instead, if at all, it should go to `this.page`.


valorkin:
Thanks, looks good to me

PR #43, opened 2015-12-10
comp(typeahead): return selected object

miqmago:
Return selected object (in case of object sources) instead of selected string

### DIFF
--- a/components/typeahead/typeahead.ts
+++ b/components/typeahead/typeahead.ts
@@ -151,7 +151,7 @@ export class TypeaheadContainer {
     if (typeof item === 'object' && this._field) {
       itemStr = item[this._field].toString();
     } else {
-        itemStr = item.toString();
+      itemStr = item.toString();
     }
     // Replaces the capture string with a the same string inside of a "strong" tag
     return query ? itemStr.replace(new RegExp(this.escapeRegexp(query), 'gi'), '<strong>$&</strong>') : itemStr;
--- a/components/typeahead/typeahead.ts
+++ b/components/typeahead/typeahead.ts
@@ -404,8 +404,14 @@ export class Typeahead implements OnInit {
   }
 
   public changeModel(value:any) {
-    this.cd.viewToModelUpdate(value);
-    setProperty(this.renderer, this.element, 'value', value);
+    var valueStr:string;
+    if (typeof value === 'object' && this.field) {
+      valueStr = value[this.field].toString();
+    } else {
+      valueStr = value.toString();
+    }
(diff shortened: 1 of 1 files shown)


components/typeahead/typeahead.ts
```diff
@@ -141,9 +146,15 @@ export class TypeaheadContainer {
     return queryToEscape.replace(/([.?*+^$[\]\\(){}|-])/g, '\\$1');
   }
 
-  private hightlight(item:string, query:string) {
+  private hightlight(item:any, query:string) {
+    var itemStr:string;
```
valorkin: @miqmago I am really against `if else` statements
can we simplify to

``` js
var itemStr = (typeof item === 'object' && this._field ? item[this._field] : item).toString()
```

components/typeahead/typeahead.ts
```diff
@@ -393,12 +404,18 @@ export class Typeahead implements OnInit {
   }
 
   public changeModel(value:any) {
-    this.cd.viewToModelUpdate(value);
-    setProperty(this.renderer, this.element, 'value', value);
+    var valueStr:string;
```
valorkin: same here

miqmago:
Done.

I've also completely rewritten some parts, in order to achive "google style" matching, implementing "latinize" to match `súper` with `super` (and vice versa) and break words with spaces (`"exact phrase" here match` would match with `match exact phrase here` but not with `phrase here exact match` - kind of "google style"). I've tested and the speed is fine. Also solves the problem with upper case hightlight.

I can make the commits here or wait to open a new PR if you are interested. (I prefer step by step).

valorkin:
LGTM!

AlonsoK28:
> Done.
> 
> I've also completely rewritten some parts, in order to achive "google style" matching, implementing "latinize" to match `súper` with `super` (and vice versa) and break words with spaces (`"exact phrase" here match` would match with `match exact phrase here` but not with `phrase here exact match` - kind of "google style"). I've tested and the speed is fine. Also solves the problem with upper case hightlight.
> 
> I can make the commits here or wait to open a new PR if you are interested. (I prefer step by step).

Excellent

How can I use TypeHeadUtils to search inside an literal object using your Algorithms (for example Latinize and others) and not to be rewrite a completly new code from scratch 💯 💯💯💯💯💯


For example like this:

```
const search ='súper';
const typeHeadConfig = {latinize: true};
const res = myObject.typeHeadUtils.search(search, typeHeadConfig);
console.log(res) //prints súper and super inside an array with other results
```

PR #44, opened 2015-12-11
comp(typeahead): latinize and single words matching

appfeel:
- typeahead-latinize: match `súper` with `super` (and vice versa)
- typeahead-single-words: match single words with exact phrase option: `"exact phrase" here match` would match with `match exact phrase here` but not with `phrase here exact match` (kind of "google style").

Also solves the problem with upper case highlighting.


valorkin:
hey, looks good, 
I have rebase and include this changes in PR #46 
Thanks for your time :+1:

PR #45, opened 2015-12-12
Upgrade to alpha 53

ciriarte:
**This is an incomplete PR**. I've only fixed buttons, pagination after the [kebab casing migration](https://github.com/angular/angular/blob/master/modules/angular2/docs/migration/kebab-case.md), as those were the two components I'm using in my app.

A bunch of the demos are broken, mostly with NgFor and string bindings, for example:

Working on alpha.44:

``` javascript
<accordion-group heading="{{group.title}}" *ng-for="#group of groups">
    {{group.content}} <= This guy will break with group being null
</accordion-group>
```

Changing it to on this PR:

``` javascript
<accordion-group [heading]="group.title" *ngFor="#group of groups">
    group.content
</accordion-group>
```

Changing `group.title` works because it is being bound, but `group.content` is not being recognized anymore. Perhaps if you can point me in the right direction I can complete this PR.

And last, but not least: I also updated a few dependencies in package.json as they were getting old.


valorkin:
hey, 
@ciriarte thanks you for your time
you commit is rebased over latest PR's
and will be a part of #47

PR #51, opened 2015-12-17
comp(typeahead) Bug: inner-html => innerHtml (BS3)

appfeel:
In angular2 beta it is no longer supported `[inner-html]`, it should be `[innerHtml]`


valorkin:
@miqmago  Thanks!

miqmago:
@valorkin :) :+1:

Issue #92, opened 2016-01-22
chore(typeahead): inside a form causes issues


ronzeidman:
Hi,
I've encountered several issues trying to put Typeahead inside a form.
1) ngControl doesn't work (throws a "No provider for NgModel! (Typeahead -> NgModel)" exception)
2) Typeahead inside a form causes submit on enter, form submission should be prevented on enter when the choices are presented (since you just want to select the choice).
3) "autocomplete" should be "off" by default.
4) Also choices should automatically close on blur (This also happens on a non-form typeahead)
5) I think "TAB" should also choose the typeahead choice.

valorkin:
> 1) ngControl doesn't work (throws a "No provider for NgModel! (Typeahead -> NgModel)" exception)

it means you should have `ngModel` on same element with `typeahead` directive

> 2) Typeahead inside a form causes submit on enter, form submission should be prevented on enter 
> when the choices are presented (since you just want to select the choice).

can you create a plunker for it? it sounds critical

> 3) "autocomplete" should be "off" by default.

will have to check it

> 4) Also choices should automatically close on blur (This also happens on a non-form typeahead)
> 5) I think "TAB" should also choose the typeahead choice.

may be on lostfocus, but yes

ronzeidman:
1) it happens also when ngModel is present.

is there a CDN or a base plunkr for ng2-bootstrap?

ronzeidman:
I've uploaded a test in my github: https://github.com/ronzeidman/open-ng2-tests
It shows the form submission.
In addition if you just add 'ngControl="something"' (not remove anything) to the input an exception will be thrown.

I'm always amazed by your fast responses!

valorkin:
> is there a CDN or a base plunkr for ng2-bootstrap?

If I will be lucky, I will land system.js bundling with TS 1.8 on this weekends
Thanks for repo, I will take a look :)

> In addition if you just add 'ngControl="something"' (not remove anything) to the input an exception will be thrown.

hm

> I'm always amazed by your fast responses!

do my best :)

nrehm:
Anyone found a solution to this problem?
Would like to use it inside a form and cant really track down why its breaking when you have  ngModel and ngControl on the same Input.

nrehm:
Digged into it a little bit and copy / changed something from the ngModel directive and added it to typeahead.directive.ts

``` javascript
const formControlBinding = 
[...]

Issue #89, opened 2016-01-19
chore(typeahead): async loading broken with angular in dev mode


danielgerlag:
The Typeahead async example no longer works, unless you put angular in production mode.
You will get "EXCEPTION: Expression has changed after it was checked." errors

https://github.com/angular/angular/issues/6005

valorkin:
try to use `ChangeDetectionStrategy.OnPush` in your component
It should solve the problem
it is not only typeahead problem, so I will have to figure out how to handle it better
current typeahead implementation will be replaced with Observables

ps2goat:
Do you know what the timeline is for the conversion to Observables?

Dinistro:
This should be fixed in the Observable version: #584

Issue #41, opened 2015-12-08
fix(typeahead): not working for async with complex objects


genereux:
I've looked into the example for the typeahead component but it is not working if I want to use the statesComplex array instead of the states array, in the async case. Could you provide an example for this case also?

valorkin:
ok, sure

Dinistro:
This should be fixed by now.

PR #99, opened 2016-01-26
Updated: rxjs for typeahead component

buchslava:


valorkin:
#92 #89 #61 #41

koodikindral:
Any progress with that?

valorkin:
yeap, will be published

Systho:
Any news on this PR ?

components/typeahead/typeahead.directive.ts
```diff
@@ -6,13 +6,13 @@ import {
 } from 'angular2/core';
 import {NgModel}from 'angular2/common';
 
+import {Observable, Subject} from 'rxjs/Rx';
```
leon, line 9: I don't think we should import the whole rxjs library here.
It's better to import them separatly, that way we don't include more code than we need.

```
import {Observable} from 'rxjs/Observable';
import {Subject} from 'rxjs/Subject';
```

valorkin, line 9: Agreed

components/typeahead/typeahead.directive.ts
```diff
@@ -58,82 +58,96 @@ export class Typeahead implements OnInit {
...
-        // then we reset the timeout to wait for the difference
-        if (last < wait) {
-          timeout = setTimeout(later, wait - last);
-          // or if not we can null out the timer and run the latest
-        } else {
-          timeout = null;
-          func.apply(this, args);
+    this.keyUpEventEmitter
+      .debounceTime(100)
+      .subscribe(
```
Dinistro, line 78: Wouldn't it be better to use flatMap() here? Like this:

```
this.keyUpEventEmitter
  .debounceTime(100)
  .flatMap(() => this.typeahead)
  .subscribe(
    (matches:string[]) => {
      this._matches = matches.slice(0, this.typeaheadOptionsLimit);
      this.finalizeAsyncCall();
    },
    (err:any) => {
      console.error(err);
    }
  );
```

Additionally we could add the `distinctUntilChanged()`-operator to prevent multiple async calls for the same input.

valorkin:
this one most probably will not be merged
because extensive rxjs usage requires additional configuration
by ng2-bootstrap consumers

Dinistro:
RxJS is in my opinion the way to go. If the async typeahead is used, the searching will mostly use http and this returns an Observable. Observables support a ton of very useful operators that can be used in the components. 
I know that this will need some documentation for all the new consumers, but the current solution is not really practical.

leon:
@valorkin Why would it require extra configuration?

if we import the operators distinctUntilChanged and debounceTime, they will be included, and the end user doesn't need to know about them.

valorkin:
ok, need to try :)

Dinistro:
@valorkin

Dinistro:
@valorkin If you need a new pull request that works with rc.1, you can inform me. I will hopefuly have some time this weekend.

valorkin:
Right now I am fighting with modals
If I win, when I win :)
Any other pop-ups stuff should be a breath
But I will appreciate any help :)

Dinistro:
@valorkin Work is in progress. This will need some time to test all the stuff. As soon that I'm finished, I will open a new PR.

valorkin:
@Dinistro awesome! I will try to finish modals today\tomorrow 
today is my Birthday, so hard to say how much time I will have :)

Dinistro:
@valorkin Happy birthday ;)
I'm finished with the work. Should I create the PR or should I wait until you merged the popup-stuff?

valorkin:
I can give you write access so you can create a branch

Dinistro:
@valorkin Would be great. I will merge the changes from #535 into the new branch.

valorkin:
@Dinistro welcome to ng2 team :)

Dinistro:
This was merged in another PR: #584

Issue #103, opened 2016-01-28
chore(build): 1.0.1-beta.2 build:prod fails with errors


bitbay:
Fresh cloning the latest 1.0.1-beta.2 and running

```
$ git clone https://github.com/valor-software/ng2-bootstrap.git
$ cd ng2-bootstrap.git
ng2-bootstrap$ npm i
ng2-bootstrap$ npm run build:prod
```

results the following error and no built/bundled files created:

```
ng2-bootstrap$ npm run build:prod
> ng2-bootstrap@1.0.1-beta.2 build:prod /workspace/github/ng2-bootstrap
> NODE_ENV=production webpack -p --progress --color --optimize-minimize --optimize-dedupe --optimize-occurence-order

clean-webpack-plugin: /workspace/github/ng2-bootstrap/build has been removed.
 54% 6/8 build modulests-loader: Using typescript@1.7.5 and /workspace/github/ng2-bootstrap/tsconfig.json
 80% optimize chunk assetsTypeError: Invalid non-string/buffer chunk
    at validChunk (_stream_writable.js:178:14)
    at Gzip.Writable.write (_stream_writable.js:205:12)
    at Gzip.Writable.end (_stream_writable.js:433:10)
    at zlibBuffer (zlib.js:197:10)
    at Object.exports.gzip (zlib.js:123:10)
    at CompressionPlugin.<anonymous> (/workspace/github/ng2-bootstrap/node_modules/compression-webpack-plugin/index.js:68:10)
    at /workspace/github/ng2-bootstrap/node_modules/compression-webpack-plugin/node_modules/async/lib/async.js:111:13
    at Array.forEach (native)
    at _each (/workspace/github/ng2-bootstrap/node_modules/compression-webpack-plugin/node_modules/async/lib/async.js:32:24)
    at Object.async.each (/workspace/github/ng2-bootstrap/node_modules/compression-webpack-plugin/node_modules/async/lib/async.js:110:9)
    at CompressionPlugin.<anonymous> (/workspace/github/ng2-bootstrap/node_modules/compression-webpack-plugin/index.js:55:10)
    at Compilation.applyPluginsAsync (/workspace/github/ng2-bootstrap/node_modules/tapable/lib/Tapable.js:71:13)
    at Compilation.<anonymous> (/workspace/github/ng2-bootstrap/node_modules/webpack/lib/Compilation.js:572:10)
[...]

PR #106, opened 2016-01-29
Corrected systemjs-bundle config meta - now using cwd.

bitbay:
Build process now does not require the local clone of the repository to be in a folder called **ng2-bootstrap**.
Please read the commits comment for more details on the issue.


valorkin:
LGTM!
Thanks for your time :)
It will not fix other issues this system.js build
I will describe in #103

PR #110, opened 2016-01-29
Fix: bundling systemjs-builder was still reading compiled files from clone/../ng2-bootstrap folder.

bitbay:
### Issue

The bundling is reading compiled files from `../ng2-bootstrap` folder.
### How to reproduce

create a new, empty folder

```
$ mkdir parent
$ cd parent
parent$ git clone https://github.com/valor-software/ng2-bootstrap.git myclone
parent$ cd myclone
parent/myclone$ npm i
...
parent/myclone$ ./make.js
Deleted files and folders:

Bundling system.js file: ng2-bootstrap.js {}
Unhandled rejection Error on fetch for ng2-bootstrap/ng2-bootstrap at file:///parent/ng2-bootstrap/ng2-bootstrap.js
    Error: ENOENT: no such file or directory, open '/parent/ng2-bootstrap/ng2-bootstrap.js'
    at Error (native)
```

---

First, adding to `tsconfig.json` the output directory as `outDir:'./ng2-bootstrap'` will make `tsc` compile the javascript files into that folder.
With this in place, there is no need to point the `baseURL` to the parent folder any more (systemjs-builder config).

Added bonus is that the compiled js files do not scatter around the source directory (modified `.gitignore`). Nice and tidy!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,13 +15,5 @@ npm-debug.log
 /bundles
 
 # ignore incline compiling
-/demo/**/*.js
-/demo/**/*.js.map
-/demo/**/*.d.ts
-/components/**/*.js
-/components/**/*.js.map
-/components/**/*.d.ts
-ng2-bootstrap.js
-ng2-bootstrap.d.ts
-ng2-bootstrap.js.map
+/ng2-bootstrap
 /logs
--- a/make.js
+++ b/make.js
@@ -32,7 +32,7 @@ async.waterfall([
 
 function getSystemJsBundleConfig(cb) {
   let config = {
-    baseURL: '..',
+    baseURL: '.',
     transpiler: 'typescript',
     typescriptOptions: {
       module: 'cjs'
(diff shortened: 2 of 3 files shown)


bitbay:
i think with this commit finally should not matter where You clone the repository and how You name it, it doesn't depend on anything outside the repository (bundling done).
Still figuring out what the webpack part has to do though...
I can run both `npm run build:prod` and `npm run build:dev` without getting any errors on the node side.
On the other hand, with `build:dev` opening up `http:\\localhost:8080` gives runtime errors - but since i get them with the current `master` branch as well, probably it is an issue for it's own (will open up).

valorkin:
as I mentioned in another PR, this will break npm cosuming

bitbay:
Your're absolutely right, i completely forgot about `npm` use-case! I see what You mean now.
Users installing with `npm i ng2-bootstrap` as a dependency won't be able to use it in that way with other bundlers.

valorkin:
yeah, it is kinda tricky to get all onboard with same usage story

Issue #38, opened 2015-11-30
fix(datepicker): for ng2 beta


heziqiang:
I found DatePicker in the Directives menu, but nothing in detail.
Do you have any plan to add it,  or go further, add the DateTimePicker?

valorkin:
It is there but need to be fixed

amanwebdev:
Anyone working on this issue? I want to take it up and also what fixes the DatePicker component needs?

valorkin:
@amansharma0091 at least make it working :)
no one is currently working on it
I would say mental model is wrong
but as first step I will appreciate if it at least will throw any errors :)

amanwebdev:
I've updated the code to beta 1 and it's working. Will make a pull request soon.
Mental model is wrong? In what ways? What should be the correct model?

valorkin:
Awesome waiting for it! :)

PR #120, opened 2016-02-01
fix(datepicker): issue #38

amanwebdev:
I've added  the flag `"suppressImplicitAnyIndexErrors":true` to tsconfig.json as I wasn't sure how to resolve implicit any index error for the datepicker popup on `this.classMap[options.placement] = true;`


tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "noImplicitAny": true,
+    "suppressImplicitAnyIndexErrors":true,
```
valorkin, line 12: @amansharma0091 removing this while merging
I have hack interface for this use case IAttribute
use like this `return (<IAttribute>this)[key + 'Text']`

ng2-bootstrap.ts
```diff
@@ -3,7 +3,7 @@ export * from  './components/alert';
 export * from  './components/buttons';
 export * from  './components/carousel';
 export * from  './components/collapse';
-// export * from  './components/datepicker/index';
+export * from  './components/datepicker/index';
```
valorkin, line 6: @amansharma0091 while merging I have applied some export shift, you will see in my commit
I want to achieve one module per component later

valorkin:
please check PR #123 
and what is your twitter name?

PR #142, opened 2016-02-04
fix(export): all the correct directives are now properly exported

NathanWalker:
Fixed up export file to include all the correct directives plus their dependent directives.
This also provides consistency with the cli `install` tool.

@valorkin Once this is merged, please recompile and post a new version build. Thanks!

### DIFF
--- a/ng2-bootstrap.ts
+++ b/ng2-bootstrap.ts
@@ -1,17 +1,17 @@
-import {Accordion, AccordionPanel} from './components/accordion';
+import {ACCORDION_DIRECTIVES} from './components/accordion';
 import {Alert} from './components/alert';
-import {ButtonCheckbox, ButtonRadio} from './components/buttons';
-import {Carousel, Slide} from './components/carousel';
+import {BUTTON_DIRECTIVES} from './components/buttons';
+import {CAROUSEL_DIRECTIVES} from './components/carousel';
 import {Collapse} from './components/collapse';
-import {DatePicker} from './components/datepicker';
-import {Dropdown} from './components/dropdown';
-import {Pagination, Pager} from './components/pagination';
-import {Progressbar, Progress} from './components/progressbar';
+import {DATEPICKER_DIRECTIVES} from './components/datepicker';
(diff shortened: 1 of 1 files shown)


valorkin:
Looks good to me, I can publish but it is useless until I remove dependency on moment
seems typing is still an issue for everyone :)

valorkin:
@NathanWalker published

jkuri:
@valorkin any idea why would the version still be `0.53.0` when installing through `ember-cli` npm script which `angular-cli` uses. I tried installing this with just `npm install ng2-bootstrap` and it installs the latest version. My guess is that `ember-cli` npm script doesn't support versioning like *-beta.2. 
Unfortunately I cannot fix that as `angular-cli` uses `ember-cli` as base. The problem is that version `0.53.0` does not include `bundles/` and correlated script in it. When this version will be out of beta this components will be compatible and I will tried then installing again.

valorkin:
try to install `@latest`?
if not, I will remove beta in version

jkuri:
Tried `ng2-bootstrap@latest` and does not work. Looks like `ember-cli` uses `semver` that does not support prereleases. 
Yes, please remove when you will have time and notify me here.

valorkin:
thing is, semver does support this kind of things http://semver.org/ :)

jkuri:
I know, not sure about version 4.x.x which `ember-cli` uses does. I am guessing.

valorkin:
@jkuri published 1.0.3 you can try now one more time

jkuri:
thanks, some peerDependencies issue now as `angular-cli` still uses angular2-beta.0 because of some jasmine problems. I will have in mind to test this again when angular-clis dependencies will be updated. 
We will make this work as `ng2-bootstrap` is really hardcore and provides many components/directives and it would be really cool to find out working with `angular-cli` as it gives you a nice overview of what the lib provides. 
I double checked the configuration of ng2-bootstrap and everything is configured properly to be compatible with angular-cli, but I cannot be sure until I see it working. Thanks again.

valorkin:
np, keep me updated :)

PR #143, opened 2016-02-05
Master

martinmcwhorter:
Added two Gulp tasks:

`gulp prepublish` 

Run before publishing to npm. Will move the source typescript to ts/

'gulp postpublish'

Run after publishing to npm. Will reverse the prepublish.


valorkin:
thanks for your help :)

will not this delete all my working ts files?
and please squash to one commit

PR #145, opened 2016-02-05
Update datepicker-inner.ts: showWeeks

edinfazlic:
I don't think it is possible to set default boolean value in Javascript. Both: showWeeks (true and false) would result in true when SHOW_WEEKS is true.


valorkin:
`this.showWeeks = !!this.showWeeks || SHOW_WEEKS;`
didn't this should be the same?

edinfazlic:
Yes, you are right. 
Updated the code line.

valorkin:
please squash commits and give it a name:
`fix(datepicker): setting default value for SHOW_WEEKS`
and we are ready to merge

valorkin:
great! thanks for your time :)

PR #146, opened 2016-02-05
Update pagination.component.ts

krzysztofsaja:
Bugfix for emiting 4 times event 'pageChanged' then component is initialized even if page isn't changed by user.

### DIFF
--- a/components/pagination/pagination.component.ts
+++ b/components/pagination/pagination.component.ts
@@ -135,12 +135,15 @@ export class Pagination implements ControlValueAccessor, OnInit, IPaginationConf
   }
 
   public set page(value) {
+    var _previous = this._page;
     this._page = (value > this.totalPages) ? this.totalPages : (value || 1);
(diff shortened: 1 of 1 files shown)


components/pagination/pagination.component.ts
```diff
@@ -135,12 +135,15 @@ export class Pagination implements ControlValueAccessor, OnInit, IPaginationConf
   }
 
   public set page(value) {
+    var _previous = this._page;
```
valorkin, line 138: please use `const`

components/pagination/pagination.component.ts
```diff
@@ -135,12 +135,15 @@ export class Pagination implements ControlValueAccessor, OnInit, IPaginationConf
...
     this._page = (value > this.totalPages) ? this.totalPages : (value || 1);
+    if (_previous !== this._page && typeof _previous !== 'undefined') {
```
valorkin, line 140: you can use negative if tree
`if page did not changed -> return`

PR #149, opened 2016-02-06
fix(pagination): multiple triggering of pageChanged event

### LINKED ISSUE #76: chore(pagination): pageChanged emitted 3 times on init
I'm using Your pagination component and I can see that pageCahnged is emitted 3 times on component init - why?

### LINKED ISSUE #138: chore(pagination): pageChanged triggered multiple times
Hi all,

I open this issue following a question on StackOverflow. See http://stackoverflow.com/questions/35081261/ng2-bootstrap-pagination-pagechanged-triggered-multiple-times/.

We saw that the `pageChanged` event is fired three times at the initialization of the `pagination` component. We expected to have only one.

After investigating a bit, it's linked to the use of a TypeScript setter for the `page` property with a custom event. Because Angular2 fires events asynchronously (see this issue: https://github.com/angular/angular/issues/6311), setting values to the `page` property fires event after the component initialization

We saw an `inited` property but it doesn't seem to be useful in this context...

We wonder if it's a normal behavior or a bug.

Here are some proposals to only have one events fired at the component startup:

(file `node_modules/ng2-bootstrap/components/pagination/pagination.ts`):
- Update the `set page` block to trigger events only when the `inited` property is `true`:
  
  ```
  public set page(value) {
    this._page = (value > this.totalPages) ? this.totalPages : (value || 1);
  
    if (this.inited) { // <---------------
      this.pageChanged.next({
        page: this._page,
        itemsPerPage: this.itemsPerPage
      });
    }
  }
  ```
- Update the `ngOnInit` method not to set the `inited` property to true at its end:
  
  ```
  ngOnInit() {
    (...)
    //this.inited = true;
  }
  ```
- Set the `inited` property to true at the end of the first call of the `writeValue`:
  
  ```
  writeValue(value:number) {
    this.page = value;
    this.pages = this.getPages(this.page, this.totalPages);
  
    if (!this.inited) {
      this.inited = true;
    }
  }
  ```

Thanks very much for your feedback!
Thierry

valorkin:
fix #76, fix #138, closes #146

### DIFF
--- a/components/pagination/pagination.component.ts
+++ b/components/pagination/pagination.component.ts
@@ -1,11 +1,6 @@
-import {
-  Component,
-  OnInit, Input, Output,
-  ElementRef, EventEmitter,
-  Self, Renderer
-} from 'angular2/core';
-import { NgFor, NgIf, NgClass, ControlValueAccessor, NgModel } from 'angular2/common';
-import { IAttribute } from '../common';
+import {Component, OnInit, Input, Output, ElementRef, EventEmitter, Self, Renderer} from 'angular2/core';
+import {NgFor, NgIf, ControlValueAccessor, NgModel} from 'angular2/common';
+import {IAttribute} from '../common';
 
 // todo: extract base functionality classes
 // todo: expose an option to change default configuration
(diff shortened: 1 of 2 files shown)


thegris:
We are still seeing this issue in 1.0.23. Has this fix been released yet?

Thanks!

valorkin:
Released a long time ago, can you provide plunkr so I can reproduce?

thegris:
OK, looks like it is an issue with having multiple copies of the same pager in the same template. This may be working as designed and I would like to know how to only fire a single pageChanged event when all of the pagers are using the same ngModel. Any advice? Thanks!

http://plnkr.co/edit/pfw8ZvQtlajKBd1IYz8g?p=preview

PR #161, opened 2016-02-10
feat(tabs): removable tabs

vvpanchenko:
Enlarge Tabs component with possibility to remove tabs.
Demo page and Readme file, updated.

### DIFF
--- a/components/tabs/tabset.component.ts
+++ b/components/tabs/tabset.component.ts
@@ -15,6 +15,9 @@ import {Tab} from './tab.directive';
             [class.active]="tabz.active" [class.disabled]="tabz.disabled"
             (click)="tabz.active = true">
             <span [ngTransclude]="tabz.headingRef">{{tabz.heading}}</span>
+            <span [hidden]="!tabz.removable">
+              <span (click)="$event.preventDefault(); removeTab(tabz);" class="glyphicon glyphicon-remove-circle"></span>
+            </span>
           </a>
         </li>
     </ul>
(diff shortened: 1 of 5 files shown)


components/tabs/tabset.component.ts
```diff
@@ -84,12 +87,41 @@ export class Tabset implements OnInit {
...
     }
 
-    this.tabs.slice(index, 1);
+    tab.removed.emit(tab);
+    this.tabs.splice(index, 1);
+  }
+
+  private getClosestTabIndex (index:number):number {
+    let tabsLength = this.tabs.length;
+    if (tabsLength) {
```
valorkin: please invert logic

components/tabs/tabset.component.ts
```diff
@@ -84,12 +87,41 @@ export class Tabset implements OnInit {
...
+          return nextIndex;
+        }
+      }
+    }
+    return -1;
+  }
+
+  private hasAvailableTabs (index:number) {
+    var tabsLength = this.tabs.length;
+    if (tabsLength) {
```
valorkin: please invert logic

components/tabs/tabset.component.ts
vvpanchenko: Logic was not changed, meaning:
1. check "active", only if True, call "hasAvailableTabs".
2. hasAvailableTabs, at first check tabs length as it was.
3. Additional logic is to check if there is any sense to search for Tab that could be switched to active.
4. And finally only if all negative cases skipped, call "getClosestTabIndex", that will start to search in both directions left/right from current index.

So, previous logic was saved, but enlarged, is it OK ?

valorkin:
```
if (!tabsLength) {
  return -1;
}
...
```

components/tabs/tabset.component.ts
valorkin:
```
if (!tabsLength) {
  return false;
}
...
```

components/tabs/tabset.component.ts
vvpanchenko: Yes, but that case still exists in the same place
Check please, source of function [removeTab](https://github.com/vvpanchenko/ng2-bootstrap/blob/master/components/tabs/tabset.component.ts#L84)

vvpanchenko:
Updated request with new commit, check please

valorkin:
Looks good, please squash commits and force push to your local branch

vvpanchenko:
Done, Thanx for your time.

valorkin:
LGTM!

Issue #124, opened 2016-02-02
chore(dropdown): with autoClose 'outsideClick' closing when clicking inside element


jurgenvo:
Hi,

i'm working on an implementation of the Dropdown and I'm having some trouble with the autoClose property. I set the autoClose to 'outsideClick' but the dropdown is toggled when I click inside the element as well. Here is what I have:

index.html:

``` html
<div id="dropdown-x" class="btn-group" dropdown [autoClose]="'outsideClick'">
    <button type="button" class="btn btn-default" dropdownToggle>
        <span class="glyphicons glyphicons-tag"></span> {{data?.reference}}
        <span class="caret"></span>
    </button>
    <div class="dropdown-menu" dropdownMenu>
        <edit-details [details]="dataDetails" (detailsApplied)="updateDetails($event)">Details</edit-details>
    </div>
</div>
```

The template of the edit-details component has the following html:

``` html
<div id="dropdown-content" *ngIf="details">
    <form role="form">
        <div class="row">
            <div class="col-lg-6">
                <input-text [(value)]="details.name" (input)="valueChanged('name', $event)"></input-text>
                ... some more inputs
            </div>
        </div>
    </form>
</div>
```

After some debugging I found that in the closeDropdown function in dropdown.service.ts, the this.openScope.menuEl is undefined. I think that the public set dropDownMenu in dropdown.directive.ts is never called, thus the this.menuEl is never set.

Any ideas on this are very welcome. Thanks!

rluba:
It seems the reason is that the menu is never set for the dropdown. This causes [the test whether the clicked element is inside the menu](https://github.com/valor-software/ng2-bootstrap/blob/55aed837041f7a3795a2fde04b9ac913b4704408/components/dropdown/dropdown.service.ts#L52) to always be false.

rluba:
Additionally, the inclusion test [only tests for equality](https://github.com/valor-software/ng2-bootstrap/blob/55aed837041f7a3795a2fde04b9ac913b4704408/components/dropdown/dropdown.service.ts#L53) instead of [doing a proper "contains" check like the original bootstrap code](https://github.com/twbs/bootstrap/blob/a1bf344c4f041ad88acaf5b2b3777c733d3afe40/js/src/dropdown.js#L176).

PR #167, opened 2016-02-10
Implement the canonical `isOpenChange` event to allow two-way binding on `isOpen`

rluba:
This allows two-way bindings like `<… dropdown [(isOpen)]="my.isOpen">` to prevent the internal and external flags from going out-of-sync whenever the dropdown is closed automatically.

From the angular2 documentation:

> […] Angular matches [(x)] to an x input property for property binding and an xChange output property for event binding.

This is a full replacement for the `onToggle` event but you might want to keep it around for backwards compatibility reasons.


valorkin:
Nice one, same as with fix for #124
please update commit message, so I can generate change logs later

rluba:
Here you go. I’ve also updated the documentation

PR #194, opened 2016-02-15
fix(demo): including es6-shim and es6-promise

binarious:
Added es6-shim and es6-promise for browsers without es6 support.

Demo: https://dl.dropboxusercontent.com/u/19031145/ng2-bootstrap/index.html
It got rid of the error `Object doesn't support property or method` in IE11, but now it throws:

```
SyntaxError: The use of a keyword for an indentifier is invalid
```

So maybe I broke something else.

### DIFF
--- a/polyfills.ts
+++ b/polyfills.ts
@@ -1,0 +1,3 @@
+// Polyfills
+import 'es6-shim';
+import 'es6-promise';
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -44,6 +44,7 @@ var config = {
   },
 
   entry: {
+    polyfills: 'polyfills',
     angular2: [
       // Angular 2 Deps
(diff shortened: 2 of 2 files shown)


ericmartinezr:
@binarious I answered you in gitter's chatroom, but maybe I'll call your attention faster in here. For your last error message there's another shim https://github.com/angular/angular/blob/master/modules/angular2/src/testing/shims_for_IE.js#L3-L14 (see @mlaval comment https://github.com/angular/angular/issues/6501#issuecomment-179502363)

binarious:
@valorkin There are still other issues with IE as you can see in the gitter discussion. This PR could still be merged to support other browsers without ES6 support. But: I needed to replace the referenced file in `CommonsChunkPlugin` to get `polyfills` to be loaded first. I don't know if that could break something else. Just be aware :).

PR #198, opened 2016-02-16
Update daypicker.ts: text-center bs4

edinfazlic:
There is no text-center only class in BS4.
BS4 introduces text-xs-center, text-sm-center...


valorkin:
Looks and works good

PR #218, opened 2016-02-24
fix(daypicker): glyphicon arrows, disable and colspan attributes

edinfazlic:
- Bootstrap 4 dropped the Glyphicons icon font. (http://v4-alpha.getbootstrap.com/migration/) Having at least something like '<' or '>' seems better.
- No need to have two titles depending on showWeeks property just to set different colspan. If showWeeks is true then it's 6, if it's false then it's 5. And since showWeeks is boolean, true/false (1/0); 5 + showWeeks would result in wanted colspan.
- Fix when disabling month header on daypicker: maxMode doesn't exist; datePicker.maxMode does


valorkin:
Looks good!

PR #219, opened 2016-02-24
fix(datepicker-inner): activeDate

edinfazlic:
When changing view on datepicker, and going left and right, selected date (activeDate) will stay the same. 
Not emitting date on move action.

### DIFF
--- a/components/datepicker/datepicker-inner.ts
+++ b/components/datepicker/datepicker-inner.ts
@@ -61,6 +61,7 @@ export class DatePickerInner implements OnInit {
   private dateFormatter:DateFormatter = new DateFormatter();
   private uniqueId:string;
   private _activeDate:Date;
+  private selectedDate:Date;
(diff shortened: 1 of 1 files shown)


valorkin:
@edinfazlic as usual
1. commit name `fix(datepicker): ...`
2. Date.valueOf is slower than to string, or implicit to number conversion (please just in case check in jsperf)

edinfazlic:
Fixed commit message.
Regarding the valueOf, here is the jsperf: http://jsperf.com/date-clone-comparation
Since there is no the same constructor in ts as in js: Date(date:Date) we cannot use that fastest method.
The other two which are the same are getTime() and valueOf(), and valueOf() sounds cleaner as getTime() may make you think that you only need time.
I agree with 2nd top answer (http://stackoverflow.com/questions/1090815/how-to-clone-a-date-object-in-javascript)

PR #220, opened 2016-02-24
feature - custom class on datepicker

edinfazlic:
Added functionality to assign custom class to specific dates on specific modes. Currently implemented for 'day' mode. Other modes still left to be implemented.

### DIFF
--- a/components/datepicker/datepicker-inner.ts
+++ b/components/datepicker/datepicker-inner.ts
@@ -91,9 +91,8 @@ export class DatePickerInner implements OnInit {
   private onlyCurrentMonth:boolean;
   @Input()
   private shortcutPropagation:boolean;
-  // todo: change type during implementation
   @Input()
-  private customClass:any;
+  private customClass:Array<{date:Date, mode:string, clazz:string}>;
   // todo: change type during implementation
   @Input()
   private dateDisabled:any;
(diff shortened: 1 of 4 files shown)


valorkin:
@edinfazlic all good, but :)
1. commit msg `feat(datepicker): ...`
2. may be we can use `ngClass` ?
3. add reference to doc

edinfazlic:
The thing is that there is already ngClass on that element. Still class could be added using ngClass, but the condition would always be true. Correct ?
So what's the point of having ngClass with class always being true ?

valorkin:
@edinfazlic don't worry
usage of ngClass in parallel with `class` is completely fine
and predictable

valorkin:
just mentioned in doc that what ever sent to `customClass` will be passed to `ngClass` without changes
and available options can be found in `ngClass` doc + link to doc

edinfazlic:
We might have miscommunicated :)

So, I was adding custom class to the `button`. Maybe you wanted it to be added to `td`one level above?
Why I did it on the button is because it is more specific.
But the thing is that there is already class on button `class="btn btn-sm"`, and there is already ngClass as well: `[ngClass]="{'btn-secondary': !dtz.selected && !datePicker.isActive(dtz), 'btn-info': dtz.selected || !dtz.selected && datePicker.isActive(dtz), disabled: dtz.disabled}"`
So, to add a class using already existing ngClass would be a bit difficult, as there is no simple way to do it. Check this [plunker](http://plnkr.co/edit/G8FHkGisnGFp8U0m6W7F?p=preview)
That's why I used one of the simplest ways - adding class as a binding to already defined classes.

And also one other thing is: if user's customClass object gets updated, it will not update custom classes in ng2-bootstrap component as there is no actual binding. Custom classes applied to every specific button are calculated only when the component is refreshed.

valorkin:
ok, got it, hm.. need to think about it

coutin:
Hi There,

Great work in here, could we have this merged soon ?

Thx

edinfazlic:
Updated last night so it supports when custom class is not passed.

components/datepicker/datepicker-inner.ts
```diff
@@ -91,9 +91,8 @@ export class DatePickerInner implements OnInit {
   private onlyCurrentMonth:boolean;
   @Input()
   private shortcutPropagation:boolean;
-  // todo: change type during implementation
   @Input()
-  private customClass:any;
+  private customClass:Array<{date:Date, mode:string, clazz:string}>;
```
valorkin, line 95: @edinfazlic all good, but:
- in doc you are mentioning `class:string`
- in code you have `clazz:string`
  so what it should be?

edinfazlic, line 95: When I changed type from any, clazz looked more proper. I didn't touch that part of readme.md. Thanks for noticing the difference. Now I updated the readme.md too.

PR #222, opened 2016-02-25
feature - display only dates from currently displayed month

edinfazlic:
Added an attribute onlyCurrentMonth which if true will not show dates from previous and next month (which was used to make a full week of 7 days). So it will show dates only from currently displayed month.

In daypicker.ts de-sugared *ngFor to be able not to even output unneeded, empty <tr>.


valorkin:
@edinfazlic ok, could be useful but
1. commit message `feat(datepicker): ....`
2. please add reference to documentation

edinfazlic:
Fixed commit message and added reference to documentation (demo).

valorkin:
awesome, I will recheck once again all 3 commits in evening
merge and new version

PR #226, opened 2016-02-26
Fix tooltip position.

olivierceulemans:
This PR should fix the position of tooltips. 

Here is the explanation of the fix: to be able to position correctly the tooltip, the tooltip has to be "rendered" by angular and the browser (otherwise size of the tooltip is reported as (0,0). So when the left, top and display style are changed in the code, they are only effective after angular change detection takes them into account.

I also removed blanks before in the beginning of the template, they were causing issues sometimes (the tooltip on the "dynamic" word in the demo exhibit this issue).

### DIFF
--- a/components/tooltip/tooltip-container.component.ts
+++ b/components/tooltip/tooltip-container.component.ts
@@ -2,7 +2,7 @@ import {
   Component,
   OnInit, Input, HostListener,
   ElementRef, EventEmitter,
-  DynamicComponentLoader, ComponentRef, Inject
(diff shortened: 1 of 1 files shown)


valorkin:
this is awesome! thanks!
I will test and merge it, if you can squash commits and use angular commit name convention
like:
`fix(tooltip): fix positioning of tooltip container`

olivierceulemans:
Done... It should be OK now.

valorkin:
Awesome, thank you

PR #291, opened 2016-03-16
fix(typeahead) - Do not invoke async method if less than typeaheadMinLength

yannickadam:
In the typeahead component, async requests are first sent, then the result gets discarded if the number of entered characters is below typeaheadMinLength. 
This PR just moves the check before doing the matching process.


yannickadam:
Squash got me :)

PR #293, opened 2016-03-16
bug(datepicker): min date sticks to current date (fixes #217)

### LINKED ISSUE #217: bug(datepicker): min date `sticks` to current date
Chrome 46
Steps to reproduce:
1. go to bs4 version http://valor-software.com/ng2-bootstrap/index-bs4.html#datepicker
2. press `min date` button
3. press any `date` on picker

Actual result:
- selected date becomes min date
  Expected:
- should not affect min date

Could be related to #163

edinfazlic:
Additionally fixed small issue in compare() in datepicker-inner.ts

But not sure if check in daypicker.ts should be removed.

### DIFF
--- a/components/datepicker/datepicker-inner.ts
+++ b/components/datepicker/datepicker-inner.ts
@@ -181,7 +181,7 @@ export class DatePickerInner implements OnInit {
       return this.compareHandlerMonth(date1, date2);
     }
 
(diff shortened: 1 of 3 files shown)


valorkin:
this is great! thanks!

PR #295, opened 2016-03-17
fix(typeahead) - Do not invoke async method if less than typeaheadMinlength

yannickadam:
In the typeahead component, async requests are first sent, then the result gets discarded if the number of entered characters is below typeaheadMinLength. 
This PR just moves the check before doing the matching process.


valorkin:
thanks, looks good to me!
I will try it later today and merge

coutin:
Hey @valorkin, do we have a target date for the pull request of the typehead and the work from @edinfazlic on the datepicker ? 

Regards,

Charles Andre

PR #319, opened 2016-03-24
Use synchronous event emitters as a workaround for dehydrated detector issues

mjamin:
See: https://github.com/angular/angular/issues/6786 and e.g. https://github.com/justindujardin/ng2-material/commit/85cab59f4331fc0a507adbaeff34f9a961998be9

### DIFF
--- a/components/alert/alert.component.ts
+++ b/components/alert/alert.component.ts
@@ -22,7 +22,7 @@ export class Alert implements OnInit {
   @Input() public dismissible:boolean;
   @Input() public dismissOnTimeout:number;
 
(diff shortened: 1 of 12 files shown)


valorkin:
@mjamin thank you for this! I would not find it with out your help for sure! :)
One question
- could we use not true\false but maybe extract a const to common.ts, to easily turn on/off this hack?
  and it will be easy to cut it off later?

mjamin:
I'd say it's simple enough this way, but I wouldn't mind adding a const. It would have to wait til tuesday, though.

valorkin:
@mjamin to not wait, I created an issue as mention
to rollback this PR changes when ng2 issue will be solved

PR #339, opened 2016-03-29
feat(affix)

Nolanus:
The bootstrap affix component as angular2 directive in typescript. Appreciating your feedback!

### DIFF
--- a/components/affix/affix.directive.ts
+++ b/components/affix/affix.directive.ts
@@ -1,5 +1,5 @@
 import {Directive, OnInit, Input, Output, EventEmitter, ElementRef, OnDestroy, HostBinding} from 'angular2/core';
-import {positionService} from '../position';
+import {positionService, ElemPosition} from '../position';
 
 export enum AffixStatus {AFFIX, AFFIX_TOP, AFFIX_BOTTOM}
 
@@ -24,17 +24,22 @@ export class Affix implements OnInit, OnDestroy {
     private isAffixedTop:boolean = true;
     @HostBinding('class.affix-bottom')
     private isAffixedBottom:boolean = true;
+    @HostBinding('style.top.px')
+    private top:number = null;
 
     @Output()
     public affixChange:EventEmitter<AffixStatusChange> = new EventEmitter(false);
 
     private status:AffixStatus = null;
     private body:HTMLBodyElement;
-    private debouncedCheckPosition:Function = Affix.debounce(() => this.checkPosition(), 10);
+    private window:Window;
+    private pinnedOffset:number = null;
+    private debouncedCheckPosition:Function = Affix.debounce(() => this.checkPosition(), 5);
     private eventListener:Function = (ev:UIEvent) => this.debouncedCheckPosition();
 
     constructor(private el:ElementRef) {
         this.body = el.nativeElement.ownerDocument.body;
+        this.window = el.nativeElement.ownerDocument.defaultView;
     }
 
     ngOnInit() {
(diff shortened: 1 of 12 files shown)


valorkin:
This is Fckn huge! Looks really good, some things could be simplified a bit, let me get to PC.
I will check it locally

One note, now you could have difficulties with squashing to one commit because of merge commit at the end, so try to use git pull rebase when possible, it will make life easier for you :)

Nolanus:
Did you find time to investigate it? Any feedback? I tried staying pretty close to the bootstrap javascript implementation, as I think they figured out the best approach to get it working in as many configurations as possible.
Let me know when I shall create a new Pull Request with everything squashed into a single commit.

koodikindral:
What's the status of this PR? :) Would be great feature

Flood:
+1, would be an awesome improvement!

pratheekhegde:
Waiting for this. +1

rg1220:
+1 Would love to see this added.

yotixify:
Any update on this? Would be nice to have!

components/affix/affix.directive.ts
```diff
@@ -0,0 +1,174 @@
...
+        this.checkPosition();
+    }
+
+    ngOnDestroy():any {
+        this.el.nativeElement.ownerDocument.defaultView.removeEventListener('scroll', this.eventListener);
+        return undefined;
+    }
+
+    private checkPosition():void {
+        let elemPos = positionService.position(this.el.nativeElement);
```
cmpilato, line 56: This call can actually fail, though not due to problems in this pull request.  See https://github.com/valor-software/ngx-bootstrap/pull/2042

components/affix/affix.directive.ts
```diff
@@ -0,0 +1,174 @@
...
+        if (newAffixStatus === AffixStatus.AFFIX_BOTTOM) {
+            this.top = scrollHeight - nativeElemPos.height - this.affixOffsetBottom;
+        }
+    }
+
+    private getState(scrollHeight:number, nativeElemPos:ElemPosition, offsetTop:number, offsetBottom:number):AffixStatus {
+        let scrollTop:number = this.body.scrollTop; // current scroll position in pixels from top
+        let targetHeight:number = this.window.innerHeight; // Height of the window / viewport area
+
+        if (offsetTop !== null && this.status === AffixStatus.AFFIX_TOP) {
```
cmpilato, line 97: Can `offsetTop` and `offsetBottom` actually be `null`?  Seems these are only passed in as `this.affixOffsetTop` and `this.affixOffsetBottom` (making having them as function parameters kinda unnecessary ...), which means they begin life initialized as `0` or another template-provided attribute value, but never `null`.  Am I missing something?

Nolanus: It's basically a pretty-close translation of the bootstrap affix javascript code and I did not question every part of their code. Nevertheless, I agree that the function arguments aren't required.
Despite a `number`-variable could still be `null` in non-`strictNullChecks` compilation, but passing in wrong values (e.g. `affixOffsetTop`) should result in wrong behaviour, so I'm fine with that as well.

cmpilato: Thanks for the explanation.

Another thing I noticed when testing in Firefox is that `body.scrollTop` is always 0.  Digging around, I found https://stackoverflow.com/questions/28633221/document-body-scrolltop-firefox-returns-0-only-js, which suggests that you might need to expand your approach for calculating `scrollTop` (here and in `getPinnedOffset()`).

kushvarmade:
Any update on this? Do we planning to add affix to ngx-bootstrap anytime soon?

Nexeuz:
this already implemented or still not?

PR #346, opened 2016-03-30
Fixed potential error if value of typeahead is undefined. Fixes #345

### LINKED ISSUE #345: "undefined" error in typeahead
Hi,

the value for the `typeahead` property of the typeahead component might be `undefined` sometimes in my app. Currently this gives an error in `typeahead.directive.ts processMatches()`, when it tries to get the `length` of `typeahead`.

PR follows.

Thanks for this otherwise great component :)

Greetings, Marc

slintes:


valorkin:
awesome catch, thank you! :)

PR #350, opened 2016-03-30
feat(BtnAccordion): add new BtnAccordion component with FontAwesome glyphs

linusbrolin:
This is an Accordion component that more resembles the Bootstrap3 Accordion.
It has a button effect and uses glyph icons from FontAwesome, since Bootstrap4 no longer includes the fonts.


valorkin:
thanks a lot for your time and help!

Issue #298, opened 2016-03-19
chore(rating): rating does not support model-driven form


hongbo-miao:
The way in sample works well:
`<rating [(ngModel)]="rate" max="5"></rating>`

But if change to model-driven form, it will show the error: 

> EXCEPTION: No value accessor for '' in [rate in TestComponent]

```
import { Component } from 'angular2/core';
import { FormBuilder, ControlGroup, AbstractControl } from 'angular2/common';

@Component({
  selector: 'test-component',
  viewProviders: [FormBuilder],
  template: `    
    <form (ngSubmit)="submit()" [ngFormModel]="form">
      <rating [ngFormControl]="rate" max="5"></rating>
      <button type="submit">Submit</button>
    </form>
  `
})
export class TestComponent {
  form: ControlGroup;
  rate: AbstractControl;

  constructor(private _formBuilder: FormBuilder) {}

  ngOnInit() {
    this.form = this._formBuilder.group({
      'rate': [3]   // or ['3'], neither works
    });

    this.rate = this.form.controls['rate'];
  }

  submit() {}
}
```

valorkin:
Yes, it doesn't by design
What is pros of using form control over ngmodel?

mkjeff:
I think the main issue was why the component need explict inject NgModel?

``` ts
class MyComp implements ControlValueAccessor
  constructor(@Self() public cd:NgModel) { // here
    cd.valueAccessor = this;
  }
}
```

It is really unnecessary and worst case will become cyclic dependency. Best way is lets framework to take over.

You can check out the usage of  [selectValueAccessor](https://github.com/angular/angular/blob/14f0e9ada8b23e6cb4dd4cf4ce1127e9c9180476/modules/angular2/src/common/forms/directives/shared.ts#L79)/[setUpControl](https://github.com/angular/angular/blob/14f0e9ada8b23e6cb4dd4cf4ce1127e9c9180476/modules/angular2/src/common/forms/directives/shared.ts#L27) in angular.

Here is a custom select comonent base on dropdown directive, it work well with ngModel/ngControl/ngFormControl. 

``` ts
import {Component, Directive, Renderer, ElementRef, Host, Self, Optional, Input, OnInit, Provider,
  provide, forwardRef, HostListener, Inject} from 'angular2/core';
import {NG_VALUE_ACCESSOR, ControlValueAccessor, NgModel } from 'angular2/common';
import {CONST_EXPR} from 'angular2/src/facade/lang';

[...]

PR #358, opened 2016-04-01
better way to implement  ngModel-compliant component

mkjeff:
fixed directive does not support model-driven form #298


valorkin:
@mkjeff hey, guys found that adding `NgModel` to providers list of components removes requirement of custom value accessors

can you please check, will it work for you?
If so it would be nice

valorkin:
does it make sense if ngControl will be deprecated?

valorkin:
Sorry for not merging it,
but now having @angular/forms
all became much easier

check this one: https://github.com/valor-software/ng2-bootstrap/blob/development/components/typeahead/typeahead.directive.ts#L176-L177

Thanks you so much for your time and help
please join slack: https://www.hamsterpad.com/chat/ng2

Issue #363, opened 2016-04-01
fix(typeahead): Clicking an item does not raise the typeaheadOnSelect event


johan13:
Repo steps:
- Go to http://valor-software.com/ng2-bootstrap/
- Type "A" in the typeahead field labelled "Asynchronous results" (Static array works the same way.)
- Click e.g. "Alaska" in the suggestions.

Result: The suggestions list disappears, but the typeaheadOnSelect is not raised and the input field is empty.
Expected: That the typeaheadOnSelect event is raised just as it would be if I'd hit enter.

johan13:
I got the same result on Mac Chrome, Mac Firefox and Windows IE11.

rbaarsma:
+1

koodikindral:
currently a blocker

koodikindral:
Seems that #351 did break this, blur event seems to override select event callback for some reason.

slintes:
ups, sorry for introducing this bug with my blur event commit. I noticed it today myself, seems I always selected by pressing enter before...

fkowal:
Hi made a fix using a different approve. https://github.com/valor-software/ng2-bootstrap/pull/398 Which one makes more sense ?

PR #389, opened 2016-04-09
fix(typeahead): Clicking an item does not raise the typeaheadOnSelectevent

koodikindral:
Fixes: #363

### DIFF
--- a/components/typeahead/typeahead-container.component.ts
+++ b/components/typeahead/typeahead-container.component.ts
@@ -10,7 +10,8 @@ const TEMPLATE:any = {
   [Ng2BootstrapTheme.BS4]: `
   <div class="dropdown-menu"
       [ngStyle]="{top: top, left: left, display: display}"
-      style="display: block">
+      style="display: block"
+      (mouseleave)="hideMenu()">
       <a href="#"
          *ngFor="#match of matches"
          class="dropdown-item"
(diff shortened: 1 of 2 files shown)


components/typeahead/typeahead-container.component.ts
```diff
@@ -10,7 +10,8 @@ const TEMPLATE:any = {
   [Ng2BootstrapTheme.BS4]: `
   <div class="dropdown-menu"
       [ngStyle]="{top: top, left: left, display: display}"
-      style="display: block">
+      style="display: block"
+      (mouseleave)="hideMenu()">
```
valorkin, line 14: @koodikindral what if it is selected by pressing tab?

slintes:
Please also see my PR https://github.com/valor-software/ng2-bootstrap/pull/395. I keep the blur handling, and fixed the mouseclick handling by listening to the mousedown event instead click event. This this will also hide the popup for keyboard-only users (which will not work with this PR afaik).

Toktik:
#395 fixes the problem

fkowal:
Fixed using a different approach https://github.com/valor-software/ng2-bootstrap/pull/398. Which one makes more sense ?

slintes:
I like your approach even more than mine :)

valorkin:
Thanks for PR, closed in favor of #398

PR #395, opened 2016-04-11
fix(typeahead): Use mousedown event instead of click event. Fixes #363

### LINKED ISSUE #363: fix(typeahead): Clicking an item does not raise the typeaheadOnSelect event

slintes:
mousedown fires before blur, see http://stackoverflow.com/questions/10652852/jquery-fire-click-before-blur-event

### DIFF
--- a/components/typeahead/typeahead-container.component.ts
+++ b/components/typeahead/typeahead-container.component.ts
@@ -14,7 +14,7 @@ const TEMPLATE:any = {
       <a href="#"
          *ngFor="#match of matches"
          class="dropdown-item"
-         (click)="selectMatch(match, $event)"
+         (mousedown)="selectMatch(match, $event)"
          (mouseenter)="selectActive(match)"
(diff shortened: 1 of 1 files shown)


fkowal:
Fix using a different approach https://github.com/valor-software/ng2-bootstrap/pull/398

koodikindral:
Looks even better! (Y)

valorkin:
closed in favor of https://github.com/valor-software/ng2-bootstrap/pull/398

PR #396, opened 2016-04-11
fix(typeahead): Add formcontrol binding. Fixes #92 (first problem)

### LINKED ISSUE #92: chore(typeahead): inside a form causes issues

slintes:
all credits go to nrehm: https://github.com/valor-software/ng2-bootstrap/issues/92#issuecomment-198986011

### DIFF
--- a/components/typeahead/typeahead.directive.ts
+++ b/components/typeahead/typeahead.directive.ts
@@ -1,8 +1,10 @@
 import {
   Directive, Input, Output, HostListener, EventEmitter, OnInit, ElementRef,
-  Renderer, DynamicComponentLoader, ComponentRef, Provider, Injector
+  Renderer, DynamicComponentLoader, ComponentRef, Provider, Injector, forwardRef
 } from 'angular2/core';
 import {NgModel} from 'angular2/common';
+import {NgControl} from 'angular2/src/common/forms/directives';
+import {CONST_EXPR} from 'angular2/src/facade/lang';
 import {TypeaheadUtils} from './typeahead-utils';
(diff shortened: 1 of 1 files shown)


fdu-axiometrics:
can we merge this to solve the ngControl issue?

Jeetah:
This bug is really annoying since it prevents using ngFormControl with TypeAhead completely. Could someone please merge it?

ghost:
Im facing the same bug here and there is a solution written, merge please?

valorkin:
no worries it will be done

Dinistro:
@slintes There are issues in this fix. Yes, it makes using of `ngControl` possible, but it crashes if you try to use a typeahead without `ngControl`. 

```
Unhandled Promise rejection: Template parse errors:
Cannot instantiate cyclic dependency! NgModel
```

marcin992:
+1

valorkin:
Seems I found a convenient way how to support both of NgModel and NgControl
finishing with docs and datepicker popup and will get here

Dinistro:
New forms and new typeahead implementation.

PR #398, opened 2016-04-12
fix(typeahead): Hide container with timeout to allow other events to propagate. Fixes #363

### LINKED ISSUE #363: fix(typeahead): Clicking an item does not raise the typeaheadOnSelect event

fkowal:
Fixes #363

### DIFF
--- a/components/typeahead/typeahead.directive.ts
+++ b/components/typeahead/typeahead.directive.ts
@@ -102,7 +102,8 @@ export class Typeahead implements OnInit {
 
   @HostListener('blur', ['$event.target'])
   protected onBlur():void {
-    this.hide();
+    // Allow typeahead container click event to be triggered requires a timeout
+    setTimeout(this.hide.bind(this), 10);
   }
 
   @HostListener('keydown', ['$event'])


valorkin:
This is nice hack, thanks! :)

johan13:
Where did the 10 ms come from? It feels a bit arbitrary.

If it is enough to wait until the next pass in the message pump, then I think we should sleep for 1 ms. (That is what `_.defer()` from lodash or underscore does.) If we need to sleep for more, then I am not comfortable with an arbitrary 10 ms. What if the browser runs on slow hardware or is heavily loaded at the time?

valorkin:
it is more likely I should do an event queue

PR #429, opened 2016-04-17
fix(Tabset): refactored tabs to avoid flicker

jhuntoo:
I was experiencing some bad flicker, whereby adding a new tab would cause content from both the new tab and the previous tab to appear at the same time, until the `'active'` class had been removed from the old pane.

I believe the cause of the issue was due to the chained usage of `tab.active` to call `tab.active` on all the other tabs. Besides from solving the flicker issue, having `Tabset` coordinate the tabs via listening to tab events and then updating the other tabs makes more sense to me. ie. A tab should be aware of itself only, and the the tabset is aware of all tabs.

### DIFF
--- a/components/common.ts
+++ b/components/common.ts
@@ -28,3 +28,12 @@ export class NgTransclude {
     this.viewRef = _viewRef;
   }
 }
+
+export class Guid {
+  static newGuid() {
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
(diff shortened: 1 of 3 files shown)


valorkin:
hey, I got your idea, but tabs should be rewritten to use @ViewChildren stuf :)

jhuntoo:
@valorkin I'm happy to do it. Do you agree with this method - http://juristr.com/blog/2016/02/learning-ng2-creating-tab-component/ ??

valorkin:
@jhuntoo it will be so cool if you can do it :D

valorkin:
Closing it for now

PR #440, opened 2016-04-20
Not setting initDate results in current date being selected

JohanSmolders:
Before the inner datepicker sets the activeDate in ngOnInit it should check if it was already set.
Otherwise it overwrites the date set by ngModel.


valorkin:
Makes sense! Thanks!

PR #445, opened 2016-04-22
fix for #16: chore(datepicker): support null value

JohanSmolders:
Introduced support for null as default value:
- In the ngOnInit method of the datepicker-inner class the update should only emit when selectedDate is changed.
- Extra check in compare function
- activeDate() shouldn't return today when not set


valorkin:
hey, looks good, let me check
but for some odd reason it shows typeahead fix too

JohanSmolders:
Sorry about that, fixed it.
Rebased with your master instead of develop.

valorkin:
no worries, always rebase on default  branch (here it is dev) :)
merging and testing locally

PR #472, opened 2016-05-02
Update *ngFor with new RC.0 syntax

leon:


kraikill:
For daypicker.component:

template variables **#index** and **#rowz** should be transformed to **let-index** and **let-rowz**

Otherwise we getting error: Template parse errors. I checked only datepicker component. The same problem may occur in other components

leon:
I was unsure if they also should be changed, but I can change them too

PR #511, opened 2016-05-15
Remove 'z' in slide name

diegochavez:
Looks like is a typo or confusing demo wrong naming on 'slide' called 'slidez'.

### DIFF
--- a/demo/components/carousel/carousel-demo.html
+++ b/demo/components/carousel/carousel-demo.html
@@ -1,14 +1,14 @@
 <div class="row">
   <div class="col-md-6">
     <carousel [interval]="myInterval" [noWrap]="noWrapSlides">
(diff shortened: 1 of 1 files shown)


valorkin:
It is not a typo, it was a time when ng2 had a name collision
directive selector - slide
var from ngFor - slide
so it was a real mess there
there are several more things like this
in tabs for example

diegochavez:
I thought was something wrong with that naming. thanks for your hard work @valorkin

valorkin:
I need to cover it with testssssssssss

valorkin:
need to update http://valor-software.com/ng2-charts/ first :(

diegochavez:
Mate I'm looking to use angular2 with charts on a new project, let me know how can help you. 👍

valorkin:
Almost done, updating demos

PR #517, opened 2016-05-17
feat(tooltip): adds implementation to tooltipEnable

dcg:
My implementation seems a little too easy not to be done already.
If i'm missing some edge cases i appreciate feedback otherwise i would be happy to see this merged :)


valorkin:
Looks good to me! Thanks for PR

PR #549, opened 2016-05-26
Fix link to source

mik-laj:


valorkin:
I am rebasing and squashing all the time, but thx for pointing :)

PR #562, opened 2016-05-30
tooltip with html support and appendToBody

erashu212:
Hi @valorkin ,

Please review and merge this PR.

Thanks

### DIFF
--- a/components/tooltip/tooltip.directive.ts
+++ b/components/tooltip/tooltip.directive.ts
@@ -1,68 +1,81 @@
 import {
-  Directive, Input, HostListener, DynamicComponentLoader,
-  ComponentRef, Provider, ReflectiveInjector, ViewContainerRef
+Directive, Input, HostListener, DynamicComponentLoader,
+ComponentRef, Provider, ReflectiveInjector, ViewContainerRef
 } from '@angular/core';
 import {TooltipOptions} from './tooltip-options.class';
 import {TooltipContainerComponent} from './tooltip-container.component';
+import {BrowserDomAdapter} from '@angular/platform-browser/src/browser_common';
 
-@Directive({selector: '[tooltip]'})
+@Directive({ selector: '[tooltip]' })
 export class TooltipDirective {
   /* tslint:disable */
-  @Input('tooltip') public content:string;
-  @Input('tooltipPlacement') public placement:string = 'top';
-  @Input('tooltipIsOpen') public isOpen:boolean;
-  @Input('tooltipEnable') public enable:boolean = true;
-  @Input('tooltipAnimation') public animation:boolean = true;
-  @Input('tooltipAppendToBody') public appendToBody:boolean;
+  @Input('tooltip') public content: string;
+  @Input('tooltipHtml') public htmlContent: string;
+  @Input('tooltipPlacement') public placement: string = 'top';
+  @Input('tooltipIsOpen') public isOpen: boolean;
+  @Input('tooltipEnable') public enable: boolean = true;
+  @Input('tooltipAnimation') public animation: boolean = true;
+  @Input('tooltipAppendToBody') public appendToBody: boolean;
   /* tslint:enable */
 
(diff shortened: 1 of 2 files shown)


erashu212:
Hi @valorkin 

Any update on this?

likerRr:
@valorkin @erashu212 I'm not sure, but does this PR really implement `appendToBody` functionality? As I can see in [source code](https://github.com/valor-software/ng2-bootstrap/blob/48234e38f8ddcd69561cb1c0c02814566b5dc9cb/components/position.ts) now `appendToBody` just changes the way of how the position of tooltip will be calculated. But expected behavior with `appendToBody=true` is that tooltip will be attached to the `body` element instead of tooltips's parent

erashu212:
Hi @likerRr ,

Is not that working?

likerRr:
Hi, @erashu212. I haven't checked it yet, but `appendToBody` should be able to solve problem when a tooltip is placed inside overflowed container.

erashu212:
Hi @likerRr ,

Please try and if possible create a plunker so we can have look into this.

Thanks

likerRr:
@erashu212 I haven't ability to try your PR now. But [here is the plunker](http://plnkr.co/edit/k0Dv2Wl1H5bQxunbGPr3?p=preview) with the common issue with the `tooltip` (it cuts inside an overflowed container) and this should be solved with `appendToBody` enabled. Could you check if your PR fixes such situation?

erashu212:
hi @likerRr ,

Thank you for pointing out this mistake. I have created an updated version of [plnkr](http://plnkr.co/edit/ns02nXoq9K5qizx1u121?p=preview).

Thanks,

components/tooltip/tooltip.directive.ts
```diff
@@ -1,68 +1,81 @@
...
     ]);
 
     this.tooltip = this.loader
       .loadNextToLocation(TooltipContainerComponent, this.viewContainerRef, binding)
-      .then((componentRef:ComponentRef<any>) => {
-        return componentRef;
-      });
+      .then((componentRef: ComponentRef<any>) => {
+      if (this.appendToBody) {
+        let dialogElement = componentRef.location.nativeElement;
```
likerRr: > `let dialogElement = componentRef.location.nativeElement;`

Why the name of tooltip element is `dialogElement`? Unsuccessful copy-paste? :)

likerRr:
@erashu212 Nice work, now it looks much better :)

components/tooltip/tooltip.directive.ts
erashu212: Hi @likerRr , 

You are right unsuccessful copy-paste :)

likerRr:
@valorkin any updates? html content and body as container are very requested features

valorkin:
append to body is most anticipated feature with ng2
because there no right approach to it
check modals for example

I will have to tripple check how this PR works, before merging

erashu212:
@valorkin : 

Did you manage to look into this?

adrianfaciu:
Maybe it would be an idea to split this in two ? One part with html content and the other with append to body. In this way html content could be merged, and append to body could be still looked into.

likerRr:
`BrowserDomAdapter` is [no longer exported as public API](https://github.com/angular/angular/issues/8509) (it's used in this PR)

adrianfaciu:
If needed I can create a pull request with the html content part only, I'm already using this from my fork. From my point of view the sooner we add more features the better :smile:

Dinistro:
@adrianfaciu Would be great

valorkin:
Ok, some things now much easier to achieve in angular2
like get document.body in a nice way (see components helper get document)
other things is using private stuff ( like browser dom adapter) will block 
system.js users from using ng2-bootstrap
- will forbid server side rendering

Thanks you for you help and time! :)
Sorry that I can not merge it

please join slack: https://www.hamsterpad.com/chat/ng2

PR #566, opened 2016-05-31
fix(dropdown): explicitly call markForCheck()

andreialecu:
Fixes https://github.com/valor-software/ng2-bootstrap/issues/563


valorkin:
Hey, thanks for PR. Have you tried dropdowns with other change detection strategies?

andreialecu:
I only tried `CheckAlways` (the default) and `OnPush`.

But all the other cases should be handled by this PR. 

If you could make a new release with this it would be great, so I can change my dependencies back to the master here. :)

valorkin:
ok, I will do release today
with couple of modal fixes + new docs + and this one

Issue #574, opened 2016-06-02
bug(modal): if backdrop === 'static' outside click should not close modal


dave-barrineau:
Provided I'm doing this correctly, I've set the modal's config to:

`<div bimodal #myModal="bs-modal" class="modal fade" [config]="{ keyboard: false, backdrop: false }" aria-labelledby="myLargeModalLabel" aria-hidden="true"> ... </div>`

However this does not prevent the modal from closing when clicking outside of the modal.

valorkin:
oh, thanks! my bad, I forgot to add check here:
https://github.com/valor-software/ng2-bootstrap/blob/development/components/modal/modal.component.ts#L79
it should be

```
@HostListener('click', ['$event'])
  protected onClick(event:any):void {
    if (this.config.backdrop === 'static' && event.target === this.element.nativeElement) {
      this.hide(event);
    }
  }
```

and options values should be added to docs (from original bs docs)
`backdrop   boolean or the string 'static'  true    Includes a modal-backdrop element. Alternatively, specify static for a backdrop which doesn't close the modal on click.`

Nice reason to make PR :)

dave-barrineau:
Thanks!

mpetkov:
Since this is an easy fix, based on the effort tag, is it possible to expedite it?

mpetkov:
Also, the code is supposed to be:
`this.config.backdrop !== 'static'`
If it is static, it shouldn't hide the modal when clicked.

calvinKG:
Hi, 

What's the solution for this bug ?

mpetkov:
@calvinKG - Bug has been fixed, update you version to get the fix.

eRez-ebs:
hi,
ran into the exact same problem, trying to download the recent version but unsuccesfully. 
when i used '`npm install ng2-bootstrap --save`' the ng2-bootstarp version in the package.json remained 1.0.17.
when i downloaded ng2-bootstarp-develpoment.zip and replaced the files manually (either those of the modal libarary or the entire folder) i got complation errors.
in both cases the result remained the same - the model was closed when clicking on the background even when i set '`data-backdrop="static"`'

can you please help me with that
cheers,
eRez

mpetkov:
@eRez-ebs - So I am also using 1.0.17.  So just passing data-backdrop won't work since that only works for the bootstrap version.  To get it to work with this library you have to pass it via the config input like so:

```
[config]="{'backdrop':'static', 'keyboard': false}"
[...]

PR #629, opened 2016-06-20
fix(modal): don't hide on out click if backdrop === 'static'

adrienverge:
Fixes: #574


valorkin:
Awesome, thanks!

adrienverge:
You're welcome! :)

Could you release a new version, so we can benefit this fix (and others) merged since v1.0.17?

PR #630, opened 2016-06-20
fix(modal): fix typo in MODAL_DIRECTIVES

adrienverge:
It should be "DIRECTIVES" (with an "I"), not "DIRECTVES".


valorkin:
Thanks a lot!

PR #632, opened 2016-06-20
fix(build): separate async pipeline to generate non-gzipped bundles (fixes #367)

### LINKED ISSUE #367: fix(build): system.js bundles are missing source maps
For servers that do not have mod_deflate enabled, the source map for `bundles/ng2-bootstrap.min.js` cannot be loaded.

sqwk:


valorkin:
Why ?

sqwk:
Because non-gzipped map files are not generated, which fails on servers that do not support mod-deflate.  (81e16b76390e2e5eb22cb39850518a0b1d8ed11b apparently did not fix the issue.)

Also when using the angular-cli with the --prod flag, the gzipped files produce an `Unexpected token ILLEGAL` error.

valorkin:
but there min\unmin and gzip\ungzip versions of bundles

sqwk:
The unzipped .map  were not included in v1.0.17. Is there another reason why they could be missing?

valorkin:
I believe this should help

PR #664, opened 2016-06-28
tooltip: add implementation for tooltipClass

adrianfaciu:
Added simple implementation for tooltipClass. Used existing popupClass property.
Styles need to be added in a main 'style.css' or use a component with ViewEncapsulation.None.


adrianfaciu:
I've added the latest changes so this could be merged. Let me know if there are additional changes required.

valorkin:
Looks good to me, thanks!

PR #665, opened 2016-06-28
Update buttons readme

jprinaldi:
The buttons readme says that when unchecking a radio button, ngModel will be set to `null`, but that feature has yet to be implemented. Currently, ngModel is being set to `undefined`.


Dinistro:
@valorkin This can be merged if you like it, no need for a successful travis build.

valorkin:
@Dinistro thank you so much :)

PR #679, opened 2016-07-01
pager: updated demo to fix pageChanged event

adrianfaciu:
Very small change to fix the demo code for pageChanged event on pager component.


valorkin:
Thanks!

Issue #703, opened 2016-07-09
ng2-bootstrap Modal - Backdrop settings are always default.


pradeepsekar:
Even after setting backdrop as false, it is not hidden. Always backdrop seems to be set as 'static' always.

Issue #687, opened 2016-07-04
Modal-Component backdrop fade effect glitches


keliel:
When activating the modal backdrop, for a fraction of a second the background gets dark without fade effect. Then it gets back to transparent and only after that a fade effect is triggered.
Tested in Chrome, Firefox and Edge. The Problem does not occur with Bootstrap only modal.

You can compare this on both demo pages:
- [ng2-bootstrap](https://valor-software.com/ng2-bootstrap/#/modals)
- [Bootstrap](https://getbootstrap.com/javascript/#modals)

PEsteves8:
I asked about that too in the gitter channel, but got no answers. It seems to me that instead of the background's element opacity starting at 0 and move gradually to somewhere near 50%, it starts at 100% and then moves down, or something.

Dinistro:
I can confirm this, but I can't see where this is coming from. Perhaps @valorkin can help.

valorkin:
to be honest I was fighting a bit with it,
what I can say and do to fix it:
- when classes and properties are added via binding - it laggs
- when via Renderer - works fine, but code looks ugly

I am waiting for ng2-animate and try it

keliel:
Can @valorkin post the said ugly code, so we can patch our page until you are able to actually close this issue.

valorkin:
OK, can I do it nearest Monday?
I am planning to release new datepicker with pop up version

BenevidesLecontes:
@valorkin @KimBum I've fixed this with css.

MaximilianWiedemann:
@BenevidesLecontes Nice to hear. Could you please share the solution with all of us?

BenevidesLecontes:
@MaximilianWiedemann here's my css with animation
.modal-backdrop{
  transition-duration: 0.4s;
  background: rgba(55, 58, 71, 0.9);
  transition: opacity 0.3s;
}
.modal-backdrop.in{
  opacity: 1;
}
.modal .modal-content{
  -webkit-animation-duration: 0.4s;
  animation-duration: 0.4s;
  -webkit-animation-fill-mode: forwards;
  animation-fill-mode: forwards;
  border-radius: 0;
}
.modal.fade.in .modal-content{
  -webkit-animation-name: anim-open;
  animation-name: anim-open;
  -webkit-animation-timing-function: cubic-bezier(0.6,0,0.4,1);
  animation-timing-function: cubic-bezier(0.6,0,0.4,1);
}

.modal.fade .modal-content{
[...]

PR #708, opened 2016-07-10
Modified to fix issues in modal backdrop and onclick events handling

pradeepsekar:
fixes for #687 & #703 
1. This fix will resolve the issue of backdrop being displayed even when
   the user sets backdrop as false.
2. This will provide option to users to control ignore backdrop clicks.

### DIFF
--- a/components/modal/modal.component.ts
+++ b/components/modal/modal.component.ts
@@ -59,7 +59,6 @@ export class ModalDirective implements AfterViewInit, OnDestroy {
   protected _isShown:boolean = false;
 
   private isBodyOverflowing:boolean = false;
-  private ignoreBackdropClick:boolean = false;
   private originalBodyPadding:number = 0;
   private scrollbarWidth:number = 0;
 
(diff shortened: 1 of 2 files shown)


components/modal/modal.component.ts
```diff
@@ -76,8 +75,7 @@ export class ModalDirective implements AfterViewInit, OnDestroy {
 
   @HostListener('click', ['$event'])
   protected onClick(event:any):void {
-    if (this.config.backdrop !== 'static' &&
-        event.target === this.element.nativeElement) {
+    if (this.config.ignoreBackdropClick || this.config.backdrop === 'static') {
```
valorkin, line 78: @pradeepsekar seem this logic should be reversed?
if ignoreBackdropClick is true or this.config.backdrop === 'static' modal should not be hidden

pradeepsekar, line 78: Hi Dimi,

You are absolutely right. it should be reversed. I will do it.

Thanks,
Pradeep

valorkin, line 78: I am planning to make a release today, so I will merge and test it locally now :)

PR #689, opened 2016-07-05
typeahead: removed incorrect behavior on TAB

rbaarsma:
TAB should simply skip to next field as expected. Solves #686 and #490

PR #715, opened 2016-07-11
fix(typeahead): removed incorrect behavior to do ENTER behavior on TAB

### LINKED ISSUE #490: fix(typeahead): Tab does not behave as click in field at Typeahead component
Changing with Tab into the Typeahead field should work as clicking with the mouse into it.
Currently Tab triggers **always** a search - event if there is nothing in the field (which could also lead to crashes if model is still undefined).
Also Tab in Typeahead popup/dropdown does not behave like Enter/Click.

### LINKED ISSUE #686: Typeahead TAB unwanted behavior
The typeahead specifically selects the active match when pressing TAB, but I think this is simply wrong.

ENTER should select the active match (as it does)
TAB should not select anything and simply move to the next field (default behavior)

To fix this only a single line needs to change in `typeahead.directive.ts`:
145      this.container.selectActiveMatch();
==>
145      this.hide()

If you guys agree I could make a pull request, but it would be simpler for me if someone from the core team can just fix it ;)

valorkin:
... TAB should simply skip to next field as expected.

fixes #686, fixes  #490, closes #689

### DIFF
--- a/components/typeahead/typeahead.directive.ts
+++ b/components/typeahead/typeahead.directive.ts
@@ -135,15 +135,9 @@ export class TypeaheadDirective implements OnInit {
       return;
     }
 
-    // if shift + tab, close items list
-    if (e.shiftKey && e.keyCode === 9) {
+    // if tab default browser behavior will select next input field, and therefore we should close the items list
+    if (e.keyCode === 9) {
       this.hide();
-      return;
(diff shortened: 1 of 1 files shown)


valorkin:
squash and merge replaces commit author! not cool at all!

PR #723, opened 2016-07-12
fix(typeahead): Added form support

Dinistro:
Typeaheads should work in forms now.


Dinistro:
I can add a hack to fix this, but it would be really, really ugly:

`(this.control.control as FormControl).updateValue(valueStr);`

The problem is, that this.control.control is not a `FromControl` everytime

valorkin:
This one resolves issue completely

``` js
    this.ngControl.viewToModelUpdate(valueStr);
    (this.ngControl.control as FormControl).updateValue(valueStr);
```

PR #724, opened 2016-07-12
Tooltip html content

adrianfaciu:
Small change adding html content to tooltip control.


adrianfaciu:
The selector name looks ok to me.  Should I change it ? Or should we change tslint rules.

components/tooltip/tooltip.directive.ts
```diff
@@ -5,10 +5,11 @@ import {
 import {TooltipOptions} from './tooltip-options.class';
 import {TooltipContainerComponent} from './tooltip-container.component';
 
-@Directive({selector: '[tooltip]'})
+@Directive({selector: '[tooltip], [tooltipHtml]'})
```
Dinistro, line 9: Tslint fails with this one. To "fix" this, you can add 

```
/* tslint:disable */
[...]
/* tslint:enable */
```

around it. The problem is a tslint extension, but I think this will be fixed in the future.

adrianfaciu:
I've disabled tslint for the tool tip directive selector. All checks are green now.

valorkin:
ok, should work as temporary solution
thanks!

PR #729, opened 2016-07-13
Simple e2e test. Header text checks

SergeyKuryatnick:
Test checks header text from all pages
also on all pages verified appearing/disappearing of the left panel with differnt window sizes

_Problem on page CAROUSEL, test failed with tag "Time out", 
but if we use option **ignore synchonization = true** - test is passed_


e2e/LeftPanelTests.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Created by vs on 7/12/16.
+ * This test is checks left panel appearing/disappearing at differ windows sizes
+ */
+
+'use strict';
```
Dinistro: `"use strict";` is not needed in TypeScript.

Dinistro:
@SergeyKuryatnick Can you fix those lint errors?: https://travis-ci.org/valor-software/ng2-bootstrap/builds/144451979#L2092-L2100

SergeyKuryatnick:
@Dinistro Done

Issue #273, opened 2016-03-11
feat(datepicker): add date picker popup version


kambojankit:
@valorkin i am trying to use the date-picker popup, instead of date-picker (as per our needs), but i am unable to make it work yet. 
I was trying out to update the code, but could not make my changes successfully. 
can i as for your suggestions, some pointers or approach to complete what is left in the popup.

sanjayjha:
So is this resolved? If yes can I get an example to use this feature?

jhuntoo:
I really need this in my project - is there anything I can do to help this along?

kambojankit:
@jhuntoo @sanjayjha i was unable look at it due to urgent project needs. For the time being we settled with [ng2-datepicker](https://github.com/jkuri/ng2-datepicker). Hope that helps.

sanjayjha:
Any update on this?

softgandhi:
+1. I would also love to get this feature.

ghost:
+1 for me as well.

piotrmalecki:
+1

AlexandrAnash:
+1

spatlollaa:
+1

spatlollaa:
+1

dothedesign:
+1 please

malifai:
+1

danielle5:
+1

steveburkett:
+1

fuitattila:
+1

jasonkrot:
plusone

Mighnus:
I like the same feature. Is it planed?

valorkin:
yeap, just I after I will get modals working

NandanPhadke:
Thanks @valorkin. I require this for my project as well. Looking forward to using this.

valorkin:
modals is working already, so I am working on datepicker popups

sraparlatmw:
+1

boban100janovski:
+1

ElBart00:
+1

vibrant-aaronstreate:
+1

geejay:
+1

Dinistro:
Please stop post +1, this messes up our notifications...

harishrohokale:
I am looking for close datepicker after date selection or any event supported which will gives us provision to handle event?

kamikaze:
+1 I've done everything doc says for datepicker directive but it doesn't work with any property and raises an error, but without - is doesn't showup without any error

[...]

PR #733, opened 2016-07-14
Emitting event when datepicker selection is done

orkisz:
Hi @valorkin,

In my project I need to have datepicker 'popup' functionality ASAP. I decided to add an event to datepicker which fires when date selection is done (picker reached minMode). Additionally, I don't want to use ngModel two way binding since it changes model value event when selection is not yet done (I want to select month, but I'm on a year view - clicking year changes model which is unacceptable for me). In this PR, I use selectionDone event to both receive selected date and as a signal to close datepicker popup.

This can be useful in #273

### DIFF
--- a/components/datepicker/datepicker-inner.component.ts
+++ b/components/datepicker/datepicker-inner.component.ts
@@ -70,7 +70,7 @@ export class DatePickerInnerComponent implements OnInit, OnChanges {
   @Input() public dateDisabled:any;
   @Input() public initDate:Date;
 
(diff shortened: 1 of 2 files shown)


valorkin:
Ok, looks good
but just in case breaking change is coming :)

valorkin:
just in case, usually docs update is required :)

orkisz:
@valorkin I did not expect you will merge this PR so fast - of course next time I will update docs :)

PR #751, opened 2016-07-18
Feature: tooltip html content as template

adrianfaciu:
Added option to define the html content of a tooltip using a [TemplateRef](https://angular.io/docs/ts/latest/api/core/index/TemplateRef-class.html), with context, using [NgTemplateOutlet](https://angular.io/docs/ts/latest/api/common/index/NgTemplateOutlet-directive.html).


valorkin:
Awesome, thanks! I like this solution :)

adrianfaciu:
@valorkin If you think this would work, I could have a look at typeahead component and see about adding a similar way to specify a template. That would make it more usable in different use cases.

valorkin:
it should work, why not

PR #768, opened 2016-07-22
TimePicker: Disabling meridian and hiding spinners if input is disabled

### LINKED ISSUE #759: TimePicker - readonlyInput option
Hi,

I wonder if it would be possible to make the option `readonlyInput` more efficient in the `TimePicker` component. The display style is correctly apply, but we can continue to change value in the input.

Thank

adrianfaciu:
If readonlyInput is true on the time picker component, the spinners will be hidden and the meridian toggle will be disabled.

If ok, should close #759.


valorkin:
Looks good to me!

PR #824, opened 2016-08-08
Slide shouldnt enforce text alignment

smasherprog:
The it should be up to the user for specific styling


valorkin:
agreed, will merge it soon

Issue #658, opened 2016-06-28
dropdowntoggle does not close when clicking directly on an icon in Chrome


jimfunk:
This seems to only happen in Chrome and it appears to be reproducible on https://valor-software.com/ng2-bootstrap/#/dropdowns

If you click directly on the down arrow, the menu opens up, but clicking on it again does not close the menu. Clicking anywhere else inside the button does close it. This also occurs with FontAwesome icons in the toggle element.

NikitaForm:
Hi! Looks like the root of this issue is here 
https://github.com/valor-software/ng2-bootstrap/blob/development/components/dropdown/dropdown.service.ts#L47
so replacing
this.openScope.toggleEl.nativeElement === event.target
with
this.openScope.toggleEl.nativeElement.contains(event.target)
will fix this bug

PR #851, opened 2016-08-17
#658 fix for dropdowntoggle does not close when clicking on icon

NikitaForm:
small fix for drop-down component


valorkin:
nice one! Thanks!

PR #882, opened 2016-08-23
merge this PR after RC6 is relased: fix breaking changes from RC6

Martin-Luft:
a future PR for RC6


Martin-Luft:
tested with the nightly builds, will not work with RC5

PR #888, opened 2016-08-24
Typeahead: Remove unused code

AlBlanc:
When i am using ng2-bootstrap with angular2 rc5 and typescript 2.0, i am facing this issue which happens because `query` is of type `string` so the condition `typeof query === 'object'` will never be fulfilled.

`ERROR in [default] /web-app/node_modules/ng2-bootstrap/components/typeahead/typeahead-container.component.ts:149:34
Property 'length' does not exist on type 'never'.`

### DIFF
--- a/components/typeahead/typeahead-container.component.ts
+++ b/components/typeahead/typeahead-container.component.ts
@@ -145,18 +145,7 @@ export class TypeaheadContainerComponent {
     let startIdx:number;
     let tokenLen:number;
     // Replaces the capture string with the same string inside of a "strong" tag
-    if (typeof query === 'object') {
-      let queryLen:number = query.length;
-      for (let i = 0; i < queryLen; i += 1) {
-        // query[i] is already latinized and lower case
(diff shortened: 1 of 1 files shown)


Martin-Luft:
This is not true. Please look at `typeahead.directive.ts`:

```
this.container.query = this.typeaheadSingleWords
  ? TypeaheadUtils.tokenize(normalizedQuery, this.typeaheadWordDelimiters, this.typeaheadPhraseDelimiters)
  : normalizedQuery;
...
this.container.query = this.typeaheadSingleWords
  ? TypeaheadUtils.tokenize(normalizedQuery, this.typeaheadWordDelimiters, this.typeaheadPhraseDelimiters)
  : normalizedQuery;
```

And `TypeaheadUtils.tokenize()` returns `Array<string>`.

So the correct solution would be to change the type from `query:string` to `query: string | string[]` (TypeScript 2) or to  `query: any` (TypeScript 1.8).

PR #897, opened 2016-08-26
fix(doc): Drop now-useless BS_VIEW_PROVIDERS instructions

adrienverge:
With the change to Angular RC5, `viewProviders` do not need to be added to components anymore. Therefore, the instruction saying "Don't forget to add view provider" are just confusing and need to be removed.


otelnov:
viewProviders yep, but we still need hack.
first one removed in https://github.com/valor-software/ng2-bootstrap/pull/864

PR #864, opened 2016-08-19
Update to rc.5

otelnov:
- Update dependencies
- Create modules for each component
- Create a global module to load all components at once
- Remove @Component({directives: [...]}) properties and add it to the component's module
- Migrate to ComponentFactoryResolver
- Replace @Component.events with @Output decorator
- Replace @Component.host with decorators
- Replace @Component.inputs with @Input decorator
- Tests refactoring
- Rewrite demo app
- Update Docs

P.S. Big thanks to @marcalj, @devoto13, @Martin-Wegner

PR #902, opened 2016-08-29
Fixed undefined event in Collapse directive

adrianfaciu:
There was an undefined event in collapse directive making it crash on Firefox.
Small change to use the state (isExpanded) for this. Should be merged with #864


otelnov:
thanks, fixed in #864

PR #909, opened 2016-09-01
use "window" instead of "Type" for the "global" substitutions

Martin-Luft:
use "window" instead of "Type" for the "global" substitutions

### DIFF
--- a/components/dropdown/dropdown-toggle.directive.ts
+++ b/components/dropdown/dropdown-toggle.directive.ts
@@ -1,13 +1,8 @@
 import {
   Directive, ElementRef, Host, HostBinding, HostListener, Input, OnInit
 } from '@angular/core';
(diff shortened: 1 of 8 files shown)


valorkin:
Please, don't remove event hacks, it is for universal renderer

Martin-Luft:
OK fixed :)

valorkin:
I have digged a bit dipper:
https://github.com/angular/angular/blob/ae75e3640a2d9eb1e897a0771d92b976c5a42c75/modules/@angular/facade/src/browser.ts

it is not so nice in in ng2 + angular universal will get working later,
may be we can drop this hack for now?

valorkin:
@Martin-Wegner fixed, tried to add your authorship but seems failed, sorry :)

Martin-Luft:
OK no problem :)

PR #925, opened 2016-09-04
style: unify html markup quotes

peterblazejewicz:
This small change unifies quotes in html markup to use
double quotes.

Thanks!


valorkin:
Nice one, thank you!:)

PR #934, opened 2016-09-06
feat(core) rename ng2-bootstrap core module to BsCoreModule

otelnov:


otelnov:
also need to fix in docs

valorkin:
We will get to it a bit later

PR #935, opened 2016-09-06
chore: update all BS references to 3.3.7

peterblazejewicz:
This commit updates all references of BS 3.0 from
3.3.\* to 3.3.7 - as 3.3.7 was introduced in #783 PR
but no references in content have been updated.

Someone should look into demo application and quickly check how it renders under 3.3.7 (I've done this, but I haven't used the project in production, so have no experience with default design).

Thanks!


valorkin:
Thank you so much!

Issue #883, opened 2016-08-23
toolTip:tooltipIsOpen


OlegMal88:
Can't get the value of tooltipIsOpen flag. Add please eventEmitter with new state of toolTip.

matthewdaniel:
i think it is implemented in such a way that you can progromatically set the isOpen value from the outside, it isn't designed to be an event. 

That being said, if that is the purpose then it isn't working that way either

adrianfaciu:
It looks like isOpen was not implemented (yet). I'll create a pull request once the feat-rc5 branch is merged into main.

hwupu:
How was it? I can't get it to work as well.

adrianfaciu:
Fixed with #939.

PR #939, opened 2016-09-07
Implementing tooltipStateChanged and exporting the directive

### LINKED ISSUE #807: can tooltip show by code control?always show?
@ciriarte @jimthedev @rkintzi @nhydock @NathanWalker

adrianfaciu:
- implemented state changed event for tooltip
- exposed the directive so it can be used in the template

Should fix #807 and #883.


valorkin:
@adrianfaciu thanks! all awesome as usual

PR #941, opened 2016-09-08
Remove moment, replace with angular2 DatePipe.  And Date.parse in demo.

Tiedye:
It has always been the intention to phase out moment, there is an alternative which is angular2's DatePipe which is always available as this is an angular2 library.  I figured we might as well use it so we can ditch the moment dependency.

In this commit I also removed moment from the docs and demo.

### DIFF
--- a/.config/bundle-system.js
+++ b/.config/bundle-system.js
@@ -59,7 +59,6 @@ function getSystemJsBundleConfig(cb) {
     memo[path.resolve(`node_modules/${currentValue}/*`)] = {build: false};
     return memo;
   }, {});
-  config.meta.moment = {build: false};
   return cb(null, config);
 }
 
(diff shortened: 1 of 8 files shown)


otelnov:
hey, @Tiedye are you in slack already? If no, please join us https://www.hamsterpad.com/chat/ng2

valorkin:
@Tiedye things is, this one will not be enough for new version of date picker we did
but I pretty like the idea
only concern ng2 DatePipe is not working properly in safari, so you have to provide polyfill for Intl

Tiedye:
@valorkin Ah I see.  Anywhere I could find the new datepicker so I could possibly try this on it?  And try getting a polyfill working if nessesary. Also thanks @otelnov I'll get right on that.

valorkin:
Join slack and dm me
Here is polifill http://npmjs.org/Intl

valorkin:
Closing it for now, we will get back to it a bit later ;)

PR #943, opened 2016-09-08
tidy up datepicker module

Martin-Luft:
try to fix some weird code which was explored in https://github.com/valor-software/ng2-bootstrap/issues/929

The `PopupContainerComponent` is not declared inside the `DatapickerModule`.
The related template contains `show-weeks` and `cupdate` which are not part of the `DatePickerComponent`. I thing it should `showWeeks` and `selectionDone`.

Hide `DatePickerInnerComponent` and `PopupContainerComponent` by removing them from `exports` because this are internal components!?


valorkin:
Better delete pop container at all
Ping me in slack I will show you a new one :)

valorkin:
I have completely removed this popup class, because it was not working properly :(

Issue #493, opened 2016-05-10
chore(build): make npm script windows compatible


ishara:
I just checkout ng2-bootstrap try to build ng2-bootstarp itself and could not run npm install.

Can u update contributor development guide to start and test?

D:\RNDA\AngulerJS\ishara\ng2-bootstrap>npm install

> node-zopfli@1.4.0 install D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! Tried to download: https://node-zopfli.s3.amazonaws.com/Release/zopfli-v1.4.0-node-v47-win32-x64.tar.gz
node-pre-gyp ERR! Pre-built binaries not found for node-zopfli@1.4.0 and node@5.10.1 (node-v47 ABI) (falling back to source compile with node-gyp)

D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli>if not defined npm_config_node_gyp (node "C:\Users\ishara\AppData\Roaming\npm\node_modules\npm\bin\node-gyp-bin\....\node_modules\node-gyp\bin\node-gyp.js" clean )  else (node "" clean )

D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli>if not defined npm_config_node_gyp (node "C:\Users\ishara\AppData\Roaming\npm\node_modules\npm\bin\node-gyp-bin\....\node_modules\node-gyp\bin\node-gyp.js" configure --fallback-to-build --module=D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli\lib\binding\node-v47-win32-x64\zopfli.node --module_name=zopfli --module_path=D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli\lib\binding\node-v47-win32-x64 )  else (node "" configure --fallback-to-build --module=D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli\lib\binding\node-v47-win32-x64\zopfli.node --module_name=zopfli --module_path=D:\RNDA\AngulerJS\ishara\ng2-bootstrap\node_modules\node-zopfli\lib\binding\node-v47-win32-x64 )

[...]

PR #953, opened 2016-09-11
Make npm script Windows compatible (#493 fix)

Igonato:
Tested on Windows 10 and Ubuntu 14.04 (x64 both)

There are numerous warning and some error messages on Windows but all seem to work at the end.

Also npm adds `./node_modules/.bin/` to PATH when it executes commands so there is no need to specify it in scripts. Feel free to revert the last commit if you want to keep it for clarity or something  ¯\_(ツ)_/¯


valorkin:
@Igonato can you create appveyor config?
so we can be sure not to break win?

valorkin:
Thanks for PR! :)

Igonato:
Happy to help. Haven't use Appveyor before, will look into it later this week. It can be tricky tho. There are errors during the installation but it seems to recover from them fine and then it was generating bundles with system-builder with no errors and produced the files but these files were empty. So I don't think there is a way to reliably test the build. Maybe relying on users reporting issues is the best option

Issue #929, opened 2016-09-05
No provider for NgModel when running tests


binarious:
I've moved to RC6. Dev and prod are running well, but when running tests I get:

```
No provider for NgModel ("[ERROR ->]<datepicker ngModel></datepicker>"): DatePickerComponent_Host@0:0
    Error: Uncaught (in promise): Error: Template parse errors:
        at resolvePromise (webpack:///~/zone.js/dist/zone.js:558:0 <- config/spec-bundle.js:89579:32)
        at resolvePromise (webpack:///~/zone.js/dist/zone.js:543:0 <- config/spec-bundle.js:89564:18)
        at webpack:///~/zone.js/dist/zone.js:591:0 <- config/spec-bundle.js:89612:18
        at ZoneDelegate.invokeTask (webpack:///~/zone.js/dist/zone.js:365:0 <- config/spec-bundle.js:89386:38)
        at ProxyZoneSpec.onInvokeTask (webpack:///~/zone.js/dist/proxy.js:130:0 <- config/spec-bundle.js:88910:44)
        at ZoneDelegate.invokeTask (webpack:///~/zone.js/dist/zone.js:364:0 <- config/spec-bundle.js:89385:43)
        at Zone.runTask (webpack:///~/zone.js/dist/zone.js:265:0 <- config/spec-bundle.js:89286:48)
        at drainMicroTaskQueue (webpack:///~/zone.js/dist/zone.js:497:0 <- config/spec-bundle.js:89518:36)
```

I'm not using the datepicker. I've just imported `Ng2BootstrapModule` in my Module.

Martin-Luft:
The Ng2BootstrapModule includes all components (e.g. Datapicker). Import only the modules you need (e.g. DropdownModule, AlarmModule...)

binarious:
Doesn't work either. It shouldn't throw an error when everything is included though.

binarious:
I removed all except the PaginationComponent which I'm using in the component I'm testing. The error changes to:

```
No provider for NgModel ("[ERROR ->]<pagination ngModel></pagination>"): PaginationComponent_Host
```

valorkin:
As far as I remember test requiring additional ngModel injection, please take a look on test for corresponding components

Martin-Luft:
The buttons, pagination, rating and typeahead modules use `ngModel` in there `selector`'s but don't import FormsModule. I'm not sure if this is necessary but I think so... Thanks for reporting this possible bug.

[...]

PR #963, opened 2016-09-13
fix(tests): "no provider" error when running tests

Igonato:
Fix for #929 

Seems like DI derps in the test environment and not able to resolve providers using the new NgModule system. For tests to pass components need to have providers field.

### DIFF
--- a/components/datepicker/datepicker.component.ts
+++ b/components/datepicker/datepicker.component.ts
@@ -31,7 +31,8 @@ import { ControlValueAccessor, NgModel } from '@angular/forms';
       <monthpicker tabindex="0"></monthpicker>
       <yearpicker tabindex="0"></yearpicker>
     </datepicker-inner>
-    `
+    `,
+  providers: [NgModel]
 })
 /* tslint:enable:component-selector-name component-selector-type */
 export class DatePickerComponent implements ControlValueAccessor {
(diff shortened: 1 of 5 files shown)


Martin-Luft:
@Igonato why is this not needed for `datepicker-popup.component` (I think this component is dead code), `button-checkbox.directive` and `button-radio.directive`? This components and directives also have the ngModel selector binding...? Or are directives not affected?

otelnov:
There is "imports" param in TestBed.configureTestingModule where you can pass FormsModule or whatever you need.

Martin-Luft:
@otelnov this is not working, please see the related issue.

Igonato:
@Martin-Wegner I don't see any datepicker-popup.component what branch are you on? 

As for the others what I did is I imported every ng2-bootstrap module and specified providers field for components that would fail the tests. Test DI doesn't fail for _every_ component nor for _every_ provider, so far it only does so with NgModel and some components. My guess is that some circular dependency needs to happen that can be resolved in the browser but not in testing.

valorkin:
@Igonato I have dropped it completely (datepicker-popup.component ) until it is working properly

Igonato:
@valorkin what was/is wrong with it? 

And what do you think about this PR? Did you review the changes, any comments?

valorkin:
as temporary solution should work, but will have to find more generic solution later

@Igonato thanks for your time! :)

PR #977, opened 2016-09-15
chore(deps): update to Angular 2.0.0

Nightapes:
update to Angular 2.0.0 Proprioception Reinforcement

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,19 +53,19 @@
     "moment": "2.15.0"
   },
   "peerDependencies": {
-    "@angular/common": "2.0.0-rc.7",
-    "@angular/compiler": "2.0.0-rc.7",
-    "@angular/core": "2.0.0-rc.7",
-    "@angular/forms": "2.0.0-rc.7"
+    "@angular/common": "2.0.0",
+    "@angular/compiler": "2.0.0",
+    "@angular/core": "2.0.0",
+    "@angular/forms": "2.0.0"
   },
   "devDependencies": {
-    "@angular/common": "2.0.0-rc.7",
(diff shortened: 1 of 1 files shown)


Martin-Luft:
But the final release is far, far away... https://github.com/angular/angular/milestones/

package.json
```diff
@@ -53,19 +53,19 @@
...
-    "@angular/platform-browser": "2.0.0-rc.7",
-    "@angular/platform-browser-dynamic": "2.0.0-rc.7",
-    "@angular/router": "3.0.0-rc.3",
+    "@angular/common": "2.0.0",
+    "@angular/compiler": "2.0.0",
+    "@angular/core": "2.0.0",
+    "@angular/forms": "2.0.0",
+    "@angular/platform-browser": "2.0.0",
+    "@angular/platform-browser-dynamic": "2.0.0",
+    "@angular/router": "3.0.0",
```
macav, line 68: would it make more sense to define it as `~2.0.0`, since there will be no more breaking changes within 2.0?

Nightapes:
As it looks, it was the official release: http://angularjs.blogspot.de/2016/09/angular2-final.html

package.json
Nightapes: Maybe we can wait till the next release, i'm not sure if there will be no more breaking changes.

valorkin: +1 to @Nightapes let it come a bit more stale

valorkin:
@Martin-Wegner ng2 is far from perfect now, but it is already much better than year ago ;)

valorkin:
@Nightapes Thanks for PR!

PR #987, opened 2016-09-15
Exclude all node_modules contents on lint

adrianfaciu:
Tried to build everything on windows again, and it works nicely. Thanks for the effort on this :+1: 

One small thing: need to manually exclude all node_module contents in the lint task, otherwise it will try to lint all the installed modules. Experienced this on windows, on linux everything was fine.


valorkin:
Thanks! :)

PR #1020, opened 2016-09-23
we should suggest to use @types/moment

Martin-Luft:
as we use TypeScript 2 and @types we should suggest to use @types/moment instead of typings


guidoffm:
It works for me but you have a typo in npm

valorkin:
Agreed! But I though moment.js is already including typings
broken one, but still enough

Issue #766, opened 2016-07-21
Add CSS class to tab in tabset


ghost:
Hi, 

I can't find a solution to add a "custom" CSS class to tab in my tabset (I want to align only one tab to the right of stack).

Do you have planned this feature in next releases?

```
<tab [disabled]="true" [ngClass]="pull-right">
    <template tabHeading>
        TITLE
    </template>
    CONTENT
</tab>
```

There is another similar issue open https://github.com/valor-software/ng2-bootstrap/issues/443 regarding the possibility to customize tabset/tab.
Thank you!

valorkin:
yeap

ghost:
Any update about this issue?

kaojo:
I am looking for same feature. 

I want to add a css class dynamically: 
`
[class.bg-info]="tabz.dragOver"
`

Confusedfish:
Just spotted kaojo's suggestion which would be better than my pr. Not sure how to implement that though

Confusedfish:
Having looked into how the `[class.classname]="binding"` works (fairly sure it's [here in the template parser](https://github.com/angular/angular/blob/f1b6c6efa1f25b725a8da6a3b91865698a7cafb4/modules/%40angular/compiler/src/template_parser/template_parser.ts#L958)) I can't see it being easy to implement the same behaviour. For now the PR should be enough to allow setting a custom class attribute on the element.

kaojo:
The PR looks solid to me. Maybe add a testcase to it? 

I don't know how to upvote (or something simular) the PR since i haven't participated in that yet. I hope it gets approved soon.

devfomichov:
Look here #1698

valorkin:
@devfomichov thanks for replying to issues :)

Issue #842, opened 2016-08-12
Need event.preventDefault in tab click event


slubowsky:
The anchors in tabs handle normal clicks like this:
``(click)="tabz.active = true"`
but clicks on remove are handled like this:
`(click)="$event.preventDefault(); removeTab(tabz);"`

Not sure under what circumstances exactly (couldn't reproduce in simple plunker) but after updating to RC 5 clicking on tabs in my application cause the app to reload (after my selected function completes). Adding $event.preventDefault(); to the anchor click function should prevent this behavior.
(Tested using jQuery

```
$('a').click(function( event ) {
  event.preventDefault();
});
```

and that fixed the issue...)

valorkin:
(click)="removeTab(tabz); return false;"
should do the trick

slubowsky:
I was referring to the ng2-bootstrap code in the tabset.component.ts template, not my code.

```
<a href class="nav-link"
            [class.active]="tabz.active" [class.disabled]="tabz.disabled"
            (click)="tabz.active = true">
            <span [ngTransclude]="tabz.headingRef">{{tabz.heading}}</span>
            <span *ngIf="tabz.removable">
              <span (click)="$event.preventDefault(); removeTab(tabz);" class="glyphicon glyphicon-remove-circle"></span>
            </span>
</a>
```

Where on remove it does $event.preventDefault(); for you but for regular clicks it doesn't. Shouldn't it?

At the time I think I was trying to get it working using $event.preventDefault(); in my click handler but that didn't work. Returning false from a handler seems to.

valorkin:
@slubowsky oh, I know where the issue could be

PR #1025, opened 2016-09-24
Issue 766

Confusedfish:
I had need to pull right a tab, this PR allows us to specify a customClass to include into the nav-item element's class attribute. This should also resolve issue #766


valorkin:
Looks good to me! @otelnov is working on AoT for ng2-bootstrap, it is almost done. Let us wait a bit and then merge this on top

Confusedfish:
In testing this I hit upon the exact same situation as issue #842 This latest commit solves it for me. The problem was highly repeatable and had me stumped for a little while. Rather than use the $event.preventDefault I just set the href to void, not sure which is better / preferred.

valorkin:
It's great, thank you!

PR #1041, opened 2016-09-27
fix(datepicker): changing model value changes the view correctly

### LINKED ISSUE #858: Datepicker UI becomes ambiguous when changing the date programmatically
If you change the bound date programmatically, the UI is very ambiguous about which date is selected.

screen shot 2016-08-18 at 1 33 46 pm

In this screenshot, the 19th is the selected date, the 20th is the last date the user has clicked on with the mouse, the 18th is the current date, and the month has been changed using the left and right arrows a few time such that the 1st appears 'pressed'. The currently selected date should be made visually distinct from all the other dates in some way, preferably with a css class that could be overwritten by apps using the datepicker component.

Example code to reproduce:

```
import { Component, OnInit } from '@angular/core';
import { FORM_DIRECTIVES } from '@angular/forms';
import { DATEPICKER_DIRECTIVES } from 'ng2-bootstrap/ng2-bootstrap';

@Component({
  selector: 'cui-example-broken-class',
  template: `
    <datepicker
          [(ngModel)]="exampleDate"
          [initDate]="exampleDate"
          [showWeeks]="false"
          minMode="day"
          maxMode="day"
          datepickerMode="day">
    </datepicker>
    <div>
    {{ exampleDate }}
    </div>
  `,
  directives: [DATEPICKER_DIRECTIVES, FORM_DIRECTIVES]
})
export class BrokenCssClassComponent implements OnInit {
  exampleDate: Date = new Date(2016, 4, 5);

  ngOnInit(): void {
    setTimeout(
        () => {
          this.exampleDate = new Date(2016, 4, 10);
        },
        5000);
  }
}
```

Remember to change the month a few times or you will not see the behavior I have described.

macav:
Resolves #858 
The problem was when the change happened from the model, the view wasn't updated properly - the change wasn't propagated to the inner component and all the attributes of date object like `selected` and `current` were not updated.

### DIFF
--- a/components/datepicker/datepicker.component.ts
+++ b/components/datepicker/datepicker.component.ts
@@ -1,4 +1,5 @@
-import { Component, EventEmitter, Input, Output, Self } from '@angular/core';
+import { Component, EventEmitter, Input, Output, Self, ViewChild } from '@angular/core';
+import { DatePickerInnerComponent } from './datepicker-inner.component';
 import { ControlValueAccessor, NgModel } from '@angular/forms';
 
 /* tslint:disable:component-selector-name component-selector-type */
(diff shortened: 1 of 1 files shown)


valorkin:
@macav thank you so much for all your help!
if you would love to join ng2 team
drop me a dm in slack https://www.hamsterpad.com/chat/ng2

Issue #853, opened 2016-08-17
Backdrop not being removed after hiding Modal


JayCallas:
I have a custom component that shows a **Modal** after clicking a button. Within the **Modal** body, the user user is presented a anchor link that routes the user away from the current view. When the user does so, the Modal is hidden/destroyed but the **Backdrop is not**.

After reading through the code, I _suspect_ that the problem is caused by the `SetTimeout()` when animation is in effect. Most likely, the objects are being destroyed before the `SetTimeout()` call happens and therefore no references are valid and the backdrop `DOM`element is left in the `DOM`.

Ideally, it would be great if within our custom component we can either change the `isAnimate` property so it propagates down to the `backdrop.instance.isAnimated` variable or the ability to call `hide()` while specifying if animation should be used for that particular call. The idea is that when showing the Modal I want to fade it onto the screen. Same for hiding the Modal _when the user hits escape or clicks the backdrop_. But when navigating away from the current view, I just want it all to immediately go away.

The only current option is to **not** show the backdrop via the `config`.

Toolo:
+1: sometimes, when I close the modal programmatically, the backdrop remains visible and I'm not able to manually close it

wombatcummings:
This issue is bugging my app as well.

Using the latest ng2 router, when I route away from the modal the bs-modal-backdrop element remains in the DOM. This can be worked around by always suppressing the backdrop using the config option.

But the bigger problem is the <body> element retains a class "modal-open" which uses overflow css to hide the scroll bars regardless of content length.  It also separately leaves padding-right css on the <body> element.

fattslug:
+1 on this. 
Hoping this is an easy fix as it makes modals _almost_ unuseable!

szh:
I am having this issue when the user navigates to a different route while a modal is open. My workaround is to put the following code in the parent component that contains the dialog:

```
ngOnDestroy(): void {
  this.modal.hide();
  $(document.body).removeClass("modal-open");
  $(".modal-backdrop").remove();
}
```

PR #1052, opened 2016-09-29
Fixed #1051. Hide modal dialog before destroying component

### LINKED ISSUE #1051: ModalDirective: Background of modal dialog is not hiding after navigation to another page
Trying to use modal directive and found an issue. Background is not hiding when you navigate to another page.
Reproducible on official demo site:
1. Navigate to http://valor-software.com/ng2-bootstrap/#/
2. Click on Modals menu
3. Click on "Large Example" button, Dialog should appear.
4. Click "Back" button in your browser
5. Observe the whole page hidden behind semi transparent dialog's background.

Also reproducible if I put routerLink in the body of my dialog to navigate to another component, so it's not linked directly to browser navigation through native buttons.

I also tried to implement OnDestroy in my component and call there dialog's hide() method, but looks like OnDestroy is not called at all in steps above, maybe I understood meaning of OnDestroy wrong.

I'll try to find a solution, but maybe someone already know a workaround.

aperevalov:
Before destroying the modal component hide() method will restore initial state of the page.

### DIFF
--- a/components/modal/modal.component.ts
+++ b/components/modal/modal.component.ts
@@ -95,6 +95,7 @@ export class ModalDirective implements AfterViewInit, OnDestroy {
   }
 
   public ngOnDestroy(): any {
+    this.hide();
     this.config = void 0;
     // this._element             = null
     // this._dialog              = null


macav:
this is duplicate of #1038 and #853

valorkin:
thanks for your PR @aperevalov

PR #1055, opened 2016-09-29
Fixed broken bindings for aria-labelby and ids

halfzebra:
Fixed broken bindings for id attributes of date-picker mode toggle button.

The rationale behind having this ARIA attribute is found in [angular-ui/bootstrap/issues/2249](https://github.com/angular-ui/bootstrap/issues/2249)

And here is the spec, which requires `aria-labelby` on the grid:
https://www.w3.org/TR/wai-aria-practices/#datepicker


valorkin:
Thanks for fix!

PR #1059, opened 2016-10-01
Adding support for wallaby.js

adrianfaciu:
Great continuous test runner. 
Will drop this here in case other people use it.

We only need two small changes:
- add wallaby webpack module
- wallaby config file

This will allow wallaby to work in any of the supported editors.
(Only tested in vs code though, can do additional testing if required)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -24,6 +24,7 @@ tsconfig.json
 tsconfig.publish.json
 tslint.json
 typedoc.js
+wallaby.js
 
 # demo build
 doc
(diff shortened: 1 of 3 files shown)


valorkin:
Wallaby js is nice! Can you please add conf file to npm ignore?

adrianfaciu:
Updated npmignore with the config file, totally forgot about it.

valorkin:
merged with https://github.com/valor-software/ng2-bootstrap/commit/cec8baecdd8a6340955f0b32e186d9e8a818f850

PR #1081, opened 2016-10-05
improved angular-cli doc

guidoffm:
Added the bootstrap@next package so there will be no need to use an external CDN


guidoffm:
I will check this again tomorrow. I think we do not need the js file but we need the css instead.

PR #1094, opened 2016-10-06
Removed parameter to prameterless method

ghost:
Having the argument on onFocus causes compiling with ngc to break


ghost:
Just  saw someone else has already done this

PR #1098, opened 2016-10-07
feat(bundle): to UMD format

lifaon74:
Quick fix from https://github.com/valor-software/ng2-bootstrap/issues/1093

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -37,6 +37,8 @@ webpack.config.js
 *.js.map
 !/bundles/*.js.map
 !*.d.ts
+!*.metadata.json
+make.js
 /components/**/*.ts
 !/components/**/*.d.ts
 
--- a/components/accordion/accordion-group.component.ts
+++ b/components/accordion/accordion-group.component.ts
@@ -4,8 +4,6 @@ import {
 
 import { AccordionComponent } from './accordion.component';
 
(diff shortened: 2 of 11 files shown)


lifaon74:
I inline source maps because I let developpers to minimize by themself files (and removing comments). It avoids to have too many files. You can change it as your conveniance.

valorkin:
Without this key mouse events, universal is not working :)

valorkin:
can you leave here only .npmignore and make.js files?
and revert all other changes?

lifaon74:
Done

valorkin:
Please check that AoT works fine,
I will play with this one :)
Thanks for quick reaction

lifaon74:
My bad, i forgot to rollback some files

valorkin:
ok, thing is
If I am planning to roll out this one
everybody will have to, replace 
`import 'ng2-bootstrap\ng2-bootstrap'` to `import 'ng2-bootstrap'`
what is actually an artifact from ng2 alfas

valorkin:
ok, testing with system.js and ng-cli (webpack)

valorkin:
ng-cli: normal and AoT versions both imports working fine ;)

valorkin:
released as @beta,
check plunker and angular2-quickstart

PR #1099, opened 2016-10-07
fix(tooltip): show & hide methods should not need any arguments

davidloiret:


valorkin:
yep, agreed
may be I will revert it back, when will add a way to prevent executing show\hide from code

PR #1104, opened 2016-10-10
fix(Typeahead): Emit `TypeaheadMatch`, not `{item: TypeaheadMatch}`

### LINKED ISSUE #1103: Typeahead: Extra nesting bug with onSelect data since v1.1.10
Hi,

`(typeaheadOnSelect)` previously emitted an `Object {item: any}` argument, from v1.1.10 it emits an `Object {item: TypeaheadMatch {item: any, value: string, header: string}}`, causing an exception in the parent component because the data is now in `event.item.item` (previously `event.item`).

One would expect either `Object {item: any}` or `TypeaheadMatch {item: any, value: string, header: string}` instead.

I'd be happy to provide additional information if needed (v1.1.9 appears to be working correctly).

adrienverge:
Previously `(typeaheadOnSelect)` emitted:

```
Object {item: any}
```

from v1.1.10 it emits:

```
Object {item: TypeaheadMatch {item: any, value: string, header: string}}
```

whereas the intended value is:

```
TypeaheadMatch {item: any, value: string, header: string}
```

This commit fixes that.

Closes: #1103

### DIFF
--- a/components/typeahead/typeahead-container.component.ts
+++ b/components/typeahead/typeahead-container.component.ts
@@ -192,9 +192,7 @@ export class TypeaheadContainerComponent {
     }
     this.parent.changeModel(value);
     setTimeout(() =>
-      this.parent.typeaheadOnSelect.emit({
-        item: value
-      }), 0
+      this.parent.typeaheadOnSelect.emit(value), 0
     );
     return false;
   }


Martin-Luft:
Thanks for your PR :) But I think you forget something. Here is my PR: https://github.com/valor-software/ng2-bootstrap/pull/1105

valorkin:
Thanks for fast fix!

PR #1105, opened 2016-10-10
fix #1103 by not nesting the TypeaheadMatch object into an object

### LINKED ISSUE #1103: Typeahead: Extra nesting bug with onSelect data since v1.1.10

Martin-Luft:
fix #1103 by not nesting the TypeaheadMatch object into an object


adrienverge:
Thanks :+1:

valorkin:
you can use it already with 
`npm install ng2-bootstrap@beta --save`

PR #1109, opened 2016-10-10
Accordion should use cards

tranlamnhatnam:
- https://github.com/twbs/bootstrap/pull/17159
- https://github.com/twbs/bootstrap/pull/20870


Martin-Luft:
The offical Bootstrap v3 example uses panels, hmmm... http://getbootstrap.com/javascript/#collapse-example-accordion

Your commit break the tests :(

OK cards are only available in v4... So we must distinguish the code between v3 and v4...

Martin-Luft:
Please have a look at https://github.com/valor-software/ng2-bootstrap/tree/PR1109

I fixed the tests... But no distinction between v3 and v4 :(

PR #1115, opened 2016-10-12
Pagination: allow setting of a custom css class on <li>

SylTi:
Make possible `<li class="btn btn-primary">` for example

### DIFF
--- a/components/pagination/pager.component.ts
+++ b/components/pagination/pager.component.ts
@@ -7,16 +7,16 @@ const pagerConfig = {
   itemsPerPage: 10,
   previousText: '« Previous',
   nextText: 'Next »',
-  classLi: '',
+  pageBtnClass: '',
   align: true
 };
 
 const PAGER_TEMPLATE = `
     <ul class="pager">
-      <li [class.disabled]="noPrevious()" [class.previous]="align" [ngClass]="{'pull-right': align}" class="{{ classLi }}">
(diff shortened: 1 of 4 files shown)


valorkin:
need to think about it a sec, 
1. classLi property doesn't describe what it does
2. `class="{{ classLi }} " need to check will it work with`[class]="classLi"`

SylTi:
1. I can change the name to something li;e `setLiClass` if you want (I'm open to suggestions).
2. [class] overwrite all existing class. That's not what we want

valorkin:
1. by meaning it seems to be `pageBtnClass` ? seems to be pretty intuitive name
2. what about `[ngClass]` ?

SylTi:
1. Fine by me I will change it to that and push the new commit.
2. It won't work for  the pager already have : `[ngClass]="{'pull-right': align}"`

valorkin:
what I am afraid of: if `pageBtnClass` will change after disable changed, it can overwrite the whole class

SylTi:
By setting the property to `btn btn-primary' : 

https://postimg.org/image/e5s4zkz2x/

Disable seems to work fine.

valorkin:
yes, now it looks much more readable,
if you are in hurry I can merge and publish it is as @beta now and cover with tests later
or you can cover it with test now

SylTi:
I will try to do the tests this weekend !

SylTi:
Is there no existing test for the pagination ? Can't find the .spec

PR #1116, opened 2016-10-12
feat(tooltip): added Tooltip delay functionality

Serginho:
I need this functionality, so I decided to implement it.

TsLint is ok.
Docs are updated.

If I forgot something, just tell me.

### DIFF
--- a/components/tooltip/tooltip.directive.ts
+++ b/components/tooltip/tooltip.directive.ts
@@ -1,5 +1,13 @@
 import {
-  ComponentRef, Directive, HostListener, Input, ReflectiveInjector, TemplateRef, ViewContainerRef, Output, EventEmitter
+  ComponentRef,
+  Directive,
+  HostListener,
+  Input,
+  ReflectiveInjector,
+  TemplateRef,
+  ViewContainerRef,
+  Output,
+  EventEmitter
 } from '@angular/core';
 
 import { TooltipContainerComponent } from './tooltip-container.component';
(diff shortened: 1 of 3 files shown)


components/tooltip/tooltip.directive.ts
```diff
@@ -43,37 +46,46 @@ export class TooltipDirective {
...
-
-    let binding = ReflectiveInjector.resolve([
-      {provide: TooltipOptions, useValue: options}
-    ]);
-
-    this.tooltip = this.componentsHelper
-      .appendNextToLocation(TooltipContainerComponent, this.viewContainerRef, binding);
 
-    this.triggerStateChanged();
+    this.delayTimeoutId = setTimeout(() => {
```
valorkin: setTimeout should be used `only` when it is passed
so can you do

```
const showTooltip = () => {...};
if (!this.delay) { showTooltip(); } else { setTimeout(showTooltip, this.delay); }
```

and than you don't need default value for delay

components/tooltip/tooltip.directive.ts
```diff
@@ -43,37 +46,46 @@ export class TooltipDirective {
...
+      this.triggerStateChanged();
+    }, this.delay);
   }
 
   // params event, target
   @HostListener('focusout')
   @HostListener('mouseleave')
   public hide():void {
+    if (this.delayTimeoutId) {
+      clearTimeout(this.delayTimeoutId);
```
valorkin, line 100: actually you can do it, like
`this.delayTimeoutId = clearTimeout(this.delayTimeoutId);`

valorkin:
Except 2 small changes looks great to me 👍

Serginho:
Do you want me to change anything else?

components/tooltip/tooltip.directive.ts
```diff
@@ -33,6 +34,8 @@ export class TooltipDirective {
   private visible:boolean = false;
   private tooltip:ComponentRef<any>;
 
+  private delayTimeoutId:NodeJS.Timer;
```
valorkin: This type can break AoT compliance, but I will have to check it manually

valorkin (changes requested):
this 3 small things

valorkin:
@Serginho oh, sorry I did not saw  this `submit review` thing before

components/tooltip/tooltip.directive.ts
Serginho: At first I used type number, but when I created the pull request, travis cancel the build because `Error: Error at /home/travis/build/valor-software/ng2-bootstrap/components/tooltip/tooltip.directive.ts:53:5: Type 'Timer' is not assignable to type 'number'.` So I changed it in the last commit.
Even WebStorm lib.d.ts uses number as a return type, but not node.

components/tooltip/tooltip.directive.ts
Serginho: This produces a lint error: `Assigned expression type void is not assignable to type NodeJS.Timer`

components/tooltip/tooltip.directive.ts
valorkin: ok, I have checked lib es6 typings, this one should be `number`

components/tooltip/tooltip.directive.ts
Serginho: Do it!

components/tooltip/tooltip.directive.ts
valorkin: yeap, sure. sorry, sometimes I forgot about TS cons :)

valorkin (approved):
Awesome! Thank you so much

components/tooltip/tooltip.directive.ts
Serginho: Ok then... What do we do with CI error? If I change to number, the build will fail.

Issue #1040, opened 2016-09-27
Accordion has forced H4 in heading


Bengreen:
The heading of panel in accordion (accordion-group.component.ts) automatically includes a H4 type heading, forcing everything inside to be a H4. Can you change this to a div and allow flexibility to change this.
If H4 is needed this can be provided by using the accordion-heading attribute.

valorkin:
Will be merged tomorrow

PR #1122, opened 2016-10-13
Update accordion

edinfazlic:
Changed Accordion panel title from h4 to div to be more WCAG compliant.


Bengreen:
Any progress on this as it would close #1040

valorkin:
applied, thank you!

PR #1155, opened 2016-10-24
Fixed modal onShow

CodingDK:
Fixed onShown event on modals so that the event not will be fired 2 times anymore. 
Before this commit it will be fired 2 times.


valorkin:
oh, it is a misprint
it should `onBeforeShow` event fired at that place
so you can prevent showing modals in last moment
same for closing, but this event was skipped for now
thanks for pointing! :)

PR #1159, opened 2016-10-24
fix(tooltip): fix `appendToBody` tooltip positioning (#1158)

### LINKED ISSUE #1158: feat(Tooltip): positioning of tooltips with `[tooltipAppendToBody]` is incorrect
I checked out the latest dev version, and it looks like the positioning of the tooltip does not always work correctly due to one argument that's not propagated. I will submit a PR shortly

DethAriel:
Closes #1158


valorkin:
Thank you!

PR #1161, opened 2016-10-24
fix(tooltip): fix delayed tooltip display (#1156)

### LINKED ISSUE #1156: feat(tooltip): Delayed tooltip seems to not be showing up
Didn't see a bug for this.

See http://valor-software.com/ng2-bootstrap/#/tooltip , and find "delayed" link-like element there. When you hover over it, the tooltip is supposed to be shown in a second, whereas in fact it never shows up (in Chrome at least). Probably the issue relates to Zones, and the `setTimeout` should be run inside the injected component Zone.

DethAriel:
Closes #1156

### DIFF
--- a/components/tooltip/tooltip.directive.ts
+++ b/components/tooltip/tooltip.directive.ts
@@ -1,4 +1,5 @@
 import {
+  ChangeDetectorRef,
   ComponentRef,
   Directive,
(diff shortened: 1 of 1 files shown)


valorkin:
Thank you!

PR #1174, opened 2016-10-26
One active slide which is changed via model - outside carouselComponent.

day01:
Changes for setting the active slide. 
With the changes developer can change model active properties on SlideComponent and it will propagate on all related slides.

With lot of fixes for whitespace and reogranize file...


valorkin:
Closing in favour of #1298
Thank you so much!

Issue #1148, opened 2016-10-21
Extending components, directives etc


DanielKucal:
Hello everyone, I probably have a feature request. I would like to extend things from this repository in my application. To give an example, I want to implement PopoverDirective basing on ng2-bootstrap TooltipDirective, like here: http://plnkr.co/edit/n5jCg3sK6VRu7fZfj7i2?p=preview
But when i try to do it in my app I get an error:

> Error: Template parse errors:
> Can't bind to 'popover' since it isn't a known property of 'div'

Above error occurs only when I extend TooltipDirective. I have "popover" set in Directive's selector, constructor() function with proper parameters, copied @Inputs() etc., so why it doesn't work?

And second thing: some of the properties are `private` - why is that? I think that they definitely should be `protected` to make really extensible components and directives.

Looking forward to hear what you think about it, thanks in advance!

valorkin:
Popover added,

PR #1186, opened 2016-10-31
extensible components

DanielKucal:
private access specifiers replaced by protected, more in #1148


valorkin:
Awesome! Perfectly correlates with ng2-bootstrap plans ;)
Thanks!

Issue #1168, opened 2016-10-26
typeahead with scroll


arins:
There is no possibility to have a long list with scroll with a typeahead component. You can add overflow-y:scroll to the ul element but that doesn't work when you use up/down arrow keys. I have made a fork and added two options two the typeahead.

typeaheadScrollable: boolean
typeaheadOptionsInScrollableView: number

if typeaheadScrollable is true the component container is scrollable and also reacts to key up/down keys.
You can also set number of items in the scrollable "view".

Is it possible to do a pull request on this when me and my colleague are done?

rdettai:
+1

arins:
yes I can look into it after this has been merged and when I have time. I agree it's a pretty important feature if you have a lot of items and can not fetch them in one go.

joebri:
+1 for this feature. Is there an ETA?

arins:
Yes I need this feature as well that's why I implemented it but it seems valor software don't have time to merge this. They have not been responding on this for a while so we went ahead and created a clone of there pacakge https://www.npmjs.com/package/symbrio-ng2-bootstrap which has the scroll. I try to update our package when they do updates and with the scroll. You can use that until they merged this.

valorkin:
@arins finaly I got here

valorkin:
@arins I am incredibly sorry, I was working with documentation >.<

arins:
Hi! I can make a new pull request. Because I saw there was a lot of changes in the master branch. It might be eaiser if I just make a new pull request.

valorkin:
@arins it would be very helpful
fork from development branch
I am merging into master only stable releases

arins:
@valorkin ok

BenElliottBV:
Hi both!

Don't suppose there's any progress on this at all? We could really use this feature!

M1strall:
Yea, we are also very interested in this feature

arins:
I have made a pull request so it's up to the repo owners to merge it. https://github.com/valor-software/ng2-bootstrap/pull/1543

Ankushi:
Any further updates on this?

M1strall:
Bump. Would be really great to support this.

Argyos:
I need this with the intensity of thousand suns. Bump

vikkivikrant:
@arins @valorkin It will be really helpful if we can add this feature.

puneetratan:
try this

```
document.addEventListener("keydown", function (e) {
[...]

PR #1195, opened 2016-11-01
scroll for typeahead

arins:
#1168 Fixed scroll for typeahead.


components/position.ts
```diff
@@ -126,6 +127,14 @@ export class PositionService {
     return (nativeEl.style as KeyAttribute)[cssProp];
   }
 
+  private get window():Window {
```
valorkin, line 130: this one can create issues with universal >.<
use https://github.com/valor-software/ng2-bootstrap/blob/development/components/utils/facade/browser.ts

components/typeahead/typeahead-container.component.ts
```diff
@@ -8,20 +8,20 @@ import { TypeaheadDirective } from './typeahead.directive';
 import { TypeaheadMatch } from './typeahead-match.class';
 
 const bs4 = `
-  <div class="dropdown-menu"
-       [ngStyle]="{top: top, left: left, display: 'block'}"
+  <div #ulElement class="dropdown-menu"
```
valorkin, line 11: html ids in reusable component? - no way
it will be a mess

valorkin (commented):
in general looks, good
but couple issue need to be fixed

Huge thanks for your time!

components/typeahead/typeahead-container.component.ts
arins: I will make a component of each dom element we want to grab. So we don't need to use element referencing by hashtag. 
When the LI and UL element are angular 2 components I can just use the code below in typehead-container.component

`@ViewChild(TypeaheadListComponent) list: TypeaheadListComponent`
`@ViewChildren(TypeaheadListItemComponent) listItems: QueryList<TypeaheadListItemComponent>`

So I will create additional components just for the UL element (`TypeaheadListComponent`)  and a component for the LI elements (`TypeaheadListItemComponent`). This way I can grab the dom elements without hashtag. What do you think about that?

components/position.ts
arins: Sorry I misunderstood you. How do you mean that it creates issues? Should I not use position.ts?

components/typeahead/typeahead-container.component.ts
valorkin: sorry sorry, my mind interpreted ng2 ids as html ids,
forgive me ;)
forget about this one

components/position.ts
arins: Is this something I need to fix? Should I not use position.ts?

PR #1211, opened 2016-11-07
use events with better support

stevenleadbeater:
mouseover and mouseout are supported across all major browsers. mouseenter is a little patchy in its support and can cause performance issues in large structures:

https://developer.mozilla.org/en/docs/Web/Events/mouseenter


valorkin:
hey, thanks for info and PR. I will merge it a bit later,
because I am in a middle of complete module restructure

valorkin:
@Betrozov I can merge now but I need your approval

valorkin:
Thanks for PR ;)

PR #1243, opened 2016-11-17
replacing unsafe-eval

### LINKED ISSUE #1242: Modal Directive uses unsafe-eval in showBackdrop function
Versions:
- ng2-boostrap 1.1.16
- angular 2.0.2

I am doing AOT compilation using ngc & rollup to create bundle.

My application is running in a CSP environment that does not allow using `eval` (or `new Function()` for that matter)

Builds fine, but I get a runtime error when I try to show modal:
image

goleary:
`new Function()` call is causing runtime error in CSP environment, and doesn't appear to be having any effect, so I've changed it
closes #1242

### DIFF
--- a/components/utils/utils.class.ts
+++ b/components/utils/utils.class.ts
@@ -2,7 +2,7 @@ import { window } from './facade/browser';
 
 export class Utils {
   public static reflow(element: any): void {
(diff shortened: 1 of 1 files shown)


valorkin:
Thanks!

PR #1263, opened 2016-11-23
fix. disable checkbox buttons

SergeyKuryatnick:
Added check for attribute:Disabled.
Fix https://github.com/valor-software/ng2-bootstrap/issues/1202


valorkin:
@SergeyKuryatnick fix should be from and to `feat-ng-cli` branch

valorkin:
merged with https://github.com/valor-software/ng2-bootstrap/commit/5e643738c5bcf8998fc8bb22c956141495d06c28

Flood:
Would be nice if this could be added to Radio Buttons as well.

sudosandwich3:
@valorkin any update for fixing this for radio buttons as well?

Issue #1270, opened 2016-11-28
Can't build with ngc+rollup - 'ModalModule' is not exported


SmallhillCZ:
Hello,
when I try to build with ngc+rollup I get a following error:

```
Export 'ComponentsHelper' is not defined by '/aot/app/app.module.ngfactory.js'                                                                                                                              
Export 'ModalModule' is not defined by '/aot/app/app.module.ngfactory.js'                                                                                                                                   
Export 'ModalModule' is not defined by '/aot/app/app.module.ngfactory.js'                                                                                                                                   
Export 'ComponentsHelper' is not defined by '/aot/app/app.module.ngfactory.js'                                                                                                                              
'ModalModule' is not exported by node_modules/ng2-bootstrap/ng2-bootstrap.js (imported by app/app.module.js).
```

and I have done everything according to https://github.com/valor-software/ng2-bootstrap/blob/development/docs/getting-started/ng-cli.md, i.e. My app.module.ts imports ModalModule. Same happens when I try to import the Ng2BootstrapModule 

also tried to config rollup with this as recommended in #1188

```
commonjs({
	include: [
		'node_modules/rxjs/**',
		'node_modules/ng2-bootstrap/'
	],
	namedExports: {
		'node_modules/ng2-bootstrap/ng2-bootstrap.js': [ 'ModalModule' ]
	}
}),
```

I am running ng2-bootstrap 1.1.16 Angular 2.2.0, angular-cli 2.2.3, rollup 0.36.4.

Without ng2-bootstrap, everything builds fine, ng-bootstrap build fine too, so I believe error is not on my side.

SmallhillCZ:
Ok, so (note the **) fixed the issue. However I'm leaving the issue open as I believe this should be either in install section or not needed.
```
commonjs({
	include: [
		'node_modules/rxjs/**',
		'node_modules/ng2-bootstrap/**'
	]
});
```

Issue #1188, opened 2016-11-01
'DropdownModule' is not exported by node_modules\ng2-bootstrap\ng2-bootstrap.js


ctrl-brk:
Trying to use rollup for treeshaking as desribed here:[http://blog.mgechev.com/2016/06/26/tree-shaking-angular2-production-build-rollup-javascript/](http://blog.mgechev.com/2016/06/26/tree-shaking-angular2-production-build-rollup-javascript/)
Getting the error `'DropdownModule' is not exported by node_modules\ng2-bootstrap\ng2-bootstrap.js` when running rollup.

Any idea what the issue could be? es2015 instead of es5?

valorkin:
Rechecked, it is exported. Try console.log(dropModule), if it is undefined, than rollup was not able to resolve ng2-bootstrap npm module

valorkin:
BTW better to use angular-cli

valorkin:
which version of ng2-bootstrap are you using BTW?
https://unpkg.com/ng2-bootstrap@1.1.16/ng2-bootstrap.js

ctrl-brk:
Found the issue. For this to work I have to use rollup-plugin-commonjs.
 Can angular-cli do tree shaking?

quiringk:
@ctrl-brk What in particular did you do when "using rollup-plugin-commonjs"? I installed rollup-plugin-commonjs and added these statements to the rollup-config.js file:

```
commonjs({
        include: 'node_modules/rxjs/**',
        namedExports: { 
          'node_modules/angular2-toaster/angular2-toaster.js': ['ToasterService'],
          'node_modules/ng2-bootstrap/ng2-bootstrap.js': ['DropdownModule']
        }
      }),
```

Which I think worked for the angular 2 toaster because it removed the same error, but am still getting the DropdownModule error with this addition.

quiringk:
Im getting this same error and don't know how to fix it with rollup-plugin-commonjs

`
'DropdownModule' is not exported by node_modules\ng2-bootstrap\ng2-bootstrap.js (imported by app\shared\navigation\navigation.module.js). For help fixing this error see https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
` 

`
[...]

PR #1273, opened 2016-11-29
Add "How to use with AoT compilation"

SmallhillCZ:
Update howto as a response to issues #1270 and #1188 with AoT compilation described at https://angular.io/docs/ts/latest/cookbook/aot-compiler.html


valorkin:
This is awesome! Thanks you!

PR #1295, opened 2016-12-05
feat(sortable): module

Le0Michine:
This PR contains implementation of d&d sortable list, at the moment it supports:
- ngModel
- drag&drop between different instances of the component
- styles for container, items, active item
- items list could be either string[] or object[] (fieldName should be provided in that case)
- animation of item insertion/removing

Unfortunately I haven't found any information about contribution rules, if it is necessary to create some demo etc. please notify me


valorkin:
Awesome!
Can you please rename to simply `sortable`?

src/index.ts
```diff
@@ -20,7 +21,7 @@ import { ComponentsHelper } from './utils/components-helper.service';
 @NgModule({
   exports: [
     AccordionModule, AlertModule, ButtonsModule, CarouselModule, CollapseModule, DatepickerModule, DropdownModule,
-    ModalModule, PaginationModule, ProgressbarModule, RatingModule, TabsModule, TimepickerModule, TooltipModule,
+    ModalModule, PaginationModule, ProgressbarModule, RatingModule, SortableListModule, TabsModule, TimepickerModule, TooltipModule,
```
valorkin: I am planning to include a lot of non bootstrap code modules here
but we should not blow up core module

src/sortable-list/index.ts
```diff
@@ -0,0 +1 @@
+export * from './sortable-list.module';
```
valorkin: this one should export all the things, to be play nice with AoT

src/sortable-list/components/sortable-list.component.ts
```diff
@@ -0,0 +1,167 @@
...
+
+import { DraggableItem } from '../models';
+import { DraggableItemService } from '../services';
+
+const nullCallback = (arg?: any): void => { return void 0; };
+
+/* tslint:disable */
+@Component({
+    selector: 'ng2-sortable-list',
+    templateUrl: './sortable-list.component.html',
```
valorkin: does angular-compiler-cli (ngc) can inline html?
I am afraid that we will have to inline html for now

valorkin (commented):
Looks good to me!
Please apply mentioned changes, tests and doc

src/index.ts
Le0Michine: You mean I need to remove it from 'exports' array?

src/sortable-list/components/sortable-list.component.ts
Le0Michine: Interesting question, I will check it today. In general don't see any problems with html inlining since it is not so big.

src/index.ts
valorkin: Any mention from this dile

src/sortable-list/components/sortable-list.component.ts
valorkin: After you will demo page you will see

Le0Michine:
@valorkin, I have added docs and tests
but there is a problem with build and I don't think that it is because of my commit, could you take a look?

Le0Michine:
@valorkin I fixed tests with pagination module (set _page = 1 by default, it was undefined by default in tests and noPrevious() method returned false because of it), so now it seems there is no blocker from CI side.
Fix in last commit 92ea666, unfortunately it has wrong comment

valorkin:
This is huge!!! Let me get to this
@SergeyKuryatnick check this out

Le0Michine:
@valorkin ping, just to remind :)

valorkin:
hey, I remember it! ;)
check new structure of src folder and how it is now published
https://unpkg.com/ng2-bootstrap@1.1.16-4/

almost ready to more not core bootstrap modules
- actually docs left...
- and publish new small version https://github.com/valor-software/ng2-bootstrap/pull/1328

valorkin:
This is huge!

valorkin:
I am actually ready to merge...
;)
drop me a note in https://ngx-slack.herokuapp.com/
and we will go through it

PR #1298, opened 2016-12-06
feat(slider): One active slide which is changed via model - outside carouselComponent.

macroorganizm:
1 commit: Linked slides together for providing only one active slide after changing
2 commit: Total rework carousel data structure
3 commit: test.spec from ng-bootstrap repo. Added functionality to setting of an active slide directly
-- Skipped tests related to keyboard events

### DIFF
--- a/demo/src/app/components/carousel/demos/carousel-demo.component.html
+++ b/demo/src/app/components/carousel/demos/carousel-demo.component.html
@@ -1,8 +1,7 @@
 <div class="row">
   <div class="col-md-4">
-    <carousel [interval]="myInterval" [noWrap]="noWrapSlides">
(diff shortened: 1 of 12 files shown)


quiringk:
Here is the code from that "carousel current slide" branch from awhile back that I have been adding to carousel component.

```
@Output() public onSlideChanged:EventEmitter<{slide:SlideComponent}> = new EventEmitter();

  public get currentSlide():SlideComponent {
    return this._currentSlide;
  }

  public set currentSlide(value:SlideComponent) {
    this._currentSlide = value;
    this.onSlideChanged.emit({slide: value});
}

protected _currentSlide:SlideComponent;
```

PR #1315, opened 2016-12-10
Make modal hack docs reflect implementation of hack in demo

tzierbock:
The implementation of the hack for setting the ApplicationRef in the modal demo was different from the documentation of how this should be done. This commit fixes this issue and brings the documentation inline with the implementation of the hack.


valorkin:
@tzierbock thank you for your help
but I am deprecating this hack with new dynamic component loader implementation:
https://github.com/valor-software/ng2-bootstrap/pull/1318

valorkin:
I hope you will like how components will look like: (this is popover)
https://github.com/valor-software/ng2-bootstrap/pull/1318/files#diff-47917cb7752d9ff4c3eddfa99fba920c
:D

PR #1362, opened 2016-12-20
Make TypeaheadMatch item public

### LINKED ISSUE #1361: AoT 1.1.16-7 TypeaheadMatch item is protected
I use this field to get the whole object but now trying to compile with AoT + 1.1.16-7 I get:

`Property 'item' is protected and only accessible within class 'TypeaheadMatch' and its subclasses.`

My code for example:

```
    public typeaheadOnSelect(e: TypeaheadMatch): void {
        this.logger.debug('typeaheadOnSelect', e);

        let selectedItem = e.item as ListItemViewModel; // My Class type

        this.members.push(selectedItem);

        this.member = '';
    }
```

philjones88:
Fixes #1361

### DIFF
--- a/src/typeahead/typeahead-match.class.ts
+++ b/src/typeahead/typeahead-match.class.ts
@@ -1,7 +1,7 @@
 
 export class TypeaheadMatch {
   public readonly value:string;
-  protected readonly item:any;
+  public readonly item:any;
   protected header:boolean;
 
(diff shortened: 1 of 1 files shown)


ajohnsonRH:
When will this be released? Thanks

valorkin:
already, v1.1.16-9

PR #1411, opened 2016-12-28
feat(popover): export popover

sallar:
Popover module is added to the base branch, but it's not exported, so it's not possible to use it in anything. So I added it to the entry so it will get exported.


valorkin:
thanks for PR! same changes with https://github.com/valor-software/ng2-bootstrap/pull/1417

valorkin:
http://valor-software.com/ng2-bootstrap/#/popover

valorkin:
here is docs

PR #1420, opened 2016-12-30
DatePicker Appointments marking

Mohendran:


valorkin:
please some description of what you are trying to achieve here
and remove console.log

PR #1433, opened 2017-01-03
play carousel on mouseup 

felixkubli:
`(mouseleave)` can't be used on a mobile devise, so a `(mouseup)` should be added.


valorkin:
@felixkubli can you please rebase on latest dev?

felixkubli:
@valorkin rebase is done.

valorkin:
awesome, thanks
seems I will have to revisit it again, (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/carousel.js#L232-L237)
but should work for now

PR #1503, opened 2017-01-16
Changed reference to BrowserModule to CommonModule instead

somombo:
Importing `BrowserModule ` (into `SortableModule`) Caused error: 
`BrowserModule has already been loaded. If you need access to common directives such as NgIf and NgFor from a lazy loaded module, import CommonModule instead.`

After trying on lazy loaded app, that had `Ng2BootstrapModule.forRoot()` imported in root App Module.

ng2-bootstrap Library should not import BrowserModule anywhere.
See [NgModule Imports section of Angular Docs](https://angular.io/docs/ts/latest/guide/ngmodule.html#!#imports) for reference


valorkin:
Thanks for pointing this out!

PR #1525, opened 2017-01-19
Fix Token , PositioningService, and other erros with mgechev-seed

sublime392:
The base index.ts was not exporting all of the folders.


valorkin:
thank you so much, published as v1.2.6

Issue #1477, opened 2017-01-11
feat(popover): click outside feature


GuskiS:
Hello, just updated to 1.1.17, everything is working amazingly, good work :) But this is not about that. Because of `ng2-bootstrap` now has popover support, I removed `ng2-popover` package as it seemed pointless to use that one anymore. The thing is, it has nice feature - `popoverCloseOnClickOutside`. It does what the name suggests - closes popover when clicked outside of it if set to true. Can we have this in `ng2-bootstrap/popover`?

valorkin:
thanks ;)
you can configure popover triggers as `triggers="click:blur"` or other focus lost event
http://valor-software.com/ng2-bootstrap/#/popover#triggers-custom

and you can configure it globally

GuskiS:
Thanks, will try it later today. Maybe there should be a place where all triggers could be listed and described(kinda)?

GuskiS:
Also, in that example is used `mouseenter:mouseleave`, how should I use triggers to achieve this kind of flow: `mouseover button shows popover, disappears when leaves button, but doesn't disappear when mouse moves from button to popover`? Basicly, lets me interact with popover?

valorkin:
mouseover:blur?

valorkin:
But I got your idea

GuskiS:
@valorkin Hmm, they are not working.
`click:blur` disappears when I click outside of popover(`correct`), disappears when I click on popover(`incorrect`)
`mouseover:blur` allows to mouseover the popover(`correct`), but doesn't disappear when I move it out of popover(`incorrect`)

valorkin:
I will need to add this functionality...

GuskiS:
Looking forward to it. Also it would be wise to add `delay` attribute to basicly delay the popover showing up and disappearing.

DanielKucal:
Implemented in https://github.com/valor-software/ng2-bootstrap/pull/1541

Klinton90:
Hopefully, fix will be available soon (thanks @DanielKucal ), but as of now, possible hack (updated to angular4 api):
```
@Component({
    ...
    host: {
//`document:click` doesn't work anymore for month- and year-picker, as contents are hidden before this handler is executed. With `mousedown` click detection works properly
        "(document:mousedown)": "onClick($event)"
    },
    template:`
<div class="form-group">
    <div 
        class="input-group"

        [popover]="bodyTemplate"
        #popover="bs-popover"
[...]

PR #1541, opened 2017-01-23
feat(popover): close on outside click

DanielKucal:
In this branch I've added new popover option `popoverCloseOnClickOutside` that when set to `true` closes popover on outside click. It was suggested in #1477, I've used the same input name as is in `pleerock/ng2-popover` to be consistent and make it easier to migrate from. 
Demo is provided. Any suggestions?
Furthemore, I was wondering what do you think about adding output event `outsideClick`? It's trivial to add now and can be helpful in some cases. Let me know, cheers :)


GuskiS:
@valorkin Mentioned(on slack) of implementing it the same as the triggers are made on button, that way making more generic than just one boolean.

src/popover/popover.directive.ts
```diff
@@ -33,6 +41,10 @@ export class PopoverDirective implements OnInit, OnDestroy {
    * Currently only supports "body".
    */
   @Input() public container: string;
+  /**
+   * Specifies if popover should be closed after clicking outside
+   */
+  @Input() public popoverCloseOnClickOutside: boolean;
```
valorkin, line 39: may be better to make it a part of config?
to not blow html?
and remove @input

src/popover/popover.directive.ts
```diff
@@ -54,10 +66,11 @@ export class PopoverDirective implements OnInit, OnDestroy {
   @Output() public onHidden: EventEmitter<any>;
 
   private _popover: ComponentLoader<PopoverContainerComponent>;
+  private _outsideClickListener: Function;
 
-  public constructor(_elementRef: ElementRef,
-                     _renderer: Renderer,
-                     _viewContainerRef: ViewContainerRef,
+  public constructor(protected _elementRef: ElementRef,
```
valorkin: please remove constructor params access modifiers

src/popover/popover.directive.ts
```diff
@@ -86,6 +99,11 @@ export class PopoverDirective implements OnInit, OnDestroy {
         placement: this.placement,
         title: this.popoverTitle
       });
+
+    if (this.popoverCloseOnClickOutside) {
```
valorkin: could it work via @HostListener?

valorkin:
looks good to me, please apply review notes ;)

src/popover/popover.directive.ts
```diff
@@ -120,4 +141,19 @@ export class PopoverDirective implements OnInit, OnDestroy {
   public ngOnDestroy(): any {
     this._popover.dispose();
   }
+
+  protected onMouseDown(target: EventTarget): void {
+    if (!this.popoverCloseOnClickOutside) return;
+    if (!this._getContainerNativeElement()) return;
+    if (this._elementRef.nativeElement === target) return;
```
valorkin: please add comment:
// if popover toggle element was clicked

src/popover/popover.directive.ts
```diff
@@ -120,4 +141,19 @@ export class PopoverDirective implements OnInit, OnDestroy {
...
+
+    if (!this._getContainerNativeElement().contains(target)) {
```
valorkin: may be change
```
// if popover popup element was clicked
if (this._getContainerNativeElement().contains(target)) return;

this.hide();
```

valorkin (changes requested):

src/popover/popover.directive.ts
DanielKucal: Um, access modifiers provide injected services as class fields. Without them I would have to declare and assign 3 variables more... E.g.

    constructor(protected _service: Service)
is an equivalent of

    protected _service: Service;
    constructor(_service: Service) {
        this._service = _service;
    }

src/popover/popover.directive.ts
DanielKucal: What in a case when someone doesn't want to set it globally? Input is more flexible, but indeed it can be also added to the config service

src/popover/popover.directive.ts
DanielKucal: Sure, but would be less efficient. Now listener is set only when this option is enabled and popover is shown, and removed after hiding.

DanielKucal:
I made some of the requested changes, but have to discuss others. 
And what about the `outsideClick` output event? Can I add it?

src/popover/popover.directive.ts
valorkin: In my plan to reduce Inputs to 5 and less per component/directive
plus you can config on level component, not globally, like in docs

src/popover/popover.directive.ts
valorkin: yep, it is. I am guilty, but I prefer implicit approach over explicit
in this case I don't what to give this binding to TS compiler

valorkin: BTW with explicit usage, you can track unused class properties

src/popover/popover.directive.ts
valorkin: ok, sounds reasonable

valorkin:
> And what about the `outsideClick` output event

why do you need it?

DanielKucal:
I thought about it again and I can achieve what I want using the `onHidden` output. Nevermind 😄

src/popover/popover.directive.ts
DanielKucal: Works with the other approach too
image

src/popover/popover.directive.ts
DanielKucal: Not good... In my case there is a big component with several injected sevices and I wouldn't want there another just for configuring a popover. Can't we leave developers the free choice how they want to use it?

GuskiS: Maybe make it a little bit shorter? Like `closeClickOutside` or `clickOutsideClose`?

GuskiS: Or, maybe just provide click outside event and let user himself to chose what to do. For example, calling `popover.close()` in that event. This way we would keep it simple, yet functional.

DanielKucal: In this case problem appears when you have more than 1 popover on the same page.

valorkin: What if, replace property with event? It will make solution more generic
so replace 
`popoverCloseOnClickOutside="true"`
with 
`(onOutsideClick)="pop.close()"` ?

DanielKucal: Okay, but wouldn't `outsideClick` be a more Angular2-way name? There is no `on` prefix in events like `click`, `dropdown` etc

DanielKucal: @valorkin, and what with setting this behavior via config? In the event-solution we would need to have both event and property to keep the functionality.

DanielKucal:
Refresh ♻️

kevinsar:
@DanielKucal any updates on this? Looks like there just needs to be some merge conflicts and test updates for this super useful PR to be merged :)

DanielKucal:
@kevinsar, I just did, however there wasn't any merge conflicts for 2 months and it hasn't been merged (they don't want more than 5 inputs here), so if you need this feature I think the best you can do for now is use `zoomsphere.ngx-bootstrap` instead. It's on the newest version (1.6.6), so all you'd have to do is replace dependency name in package.json and import statement like:

    import { Ng2BootstrapModule } from 'zoomsphere.ngx-bootstrap';

valorkin (commented):
Looks good to me, but it should work for any element created via component loader and react not only on click, but rather be configurable like targetTriggers (any ideas for name? )
At the moment tooltip, popover and dropdown use component loader, later on all dynamic components will use it.

What do you think?

AhsanAyaz:
@DanielKucal @valorkin 
Any updates on this?
Hoping to have this feature soon in the next release :)

cedricremond:
Any news for this feature ? 
Seems to be really wanted, and I tested from the zoosphere.ngx-bootstrap repo and it doesn't seems to work anymore (I found the popoverCloseOnClickOutside directive but setting it to true doesn't change anything).

valorkin:
We will do it via component loader, so all dynamic components will get this feature

Planned for nearest release

cedricremond:
@valorkin 
Nice ! Do you know when this release will be published ?

valorkin:
merged with https://github.com/valor-software/ngx-bootstrap/pull/2441
thanks for PR

andre-gois:
Is this released? Can't find it anywhere on the documentation.

valorkin:
always up to date docs can be found here: http://ngx-bootstrap.surge.sh/#/

valorkin:
I need to resolve blocking issue with build for system.js

valorkin:
except it, all is ready for v2 (

Issue #1406, opened 2016-12-27
Tooltip with delay only appears once


zkorz:
I noticed this on my page, but it also is an issue on your demo page:

![image](https://cloud.githubusercontent.com/assets/1299559/21506829/38d21b3e-cc41-11e6-8d48-ff9d74068079.png)

Hover over "delayed" for one second and the tooltip appears (as expected), but then unhover and hover over it again and it will never appear a second time.

Valery-vs:
Hi. When could we see this fix in release?

AhsanAyaz:
@valorkin when can we see in a release? 
I'm building a project for a client and it really requires the delayed effect 😄

IlyaSurmay:
fixed in #2826

PR #1546, opened 2017-01-24
fix(tooltip): Fixed delayed tooltips[#1406]

macroorganizm:
Fixes https://github.com/valor-software/ng2-bootstrap/issues/1406

### DIFF
--- a/src/tooltip/tooltip.directive.ts
+++ b/src/tooltip/tooltip.directive.ts
@@ -6,6 +6,7 @@ import { TooltipContainerComponent } from './tooltip-container.component';
 import { TooltipConfig } from './tooltip.config';
 import { ComponentLoaderFactory, ComponentLoader } from '../component-loader';
 import { OnChange } from '../utils/decorators';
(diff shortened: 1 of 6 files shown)


src/tooltip/tooltip.directive.ts
```diff
@@ -208,7 +220,24 @@ export class TooltipDirective implements OnInit, OnDestroy {
...
+        .subscribe((data) => {
+          isShowingEnabled = data.showEnabled;
+        });
+
+      this._delayTimeoutId = setTimeout(() => {
+        subscription.unsubscribe();
+        if (isShowingEnabled) {
+          this._delayTimeoutId = void 0;
+          showTooltip();
+        } else {
```
valorkin: remove `else` and use `return`

valorkin (changes requested):
and please add tests

arekzelechowski:
Should we consider this pull request rejected? What is the reason? I am thinking about using this pull request in production and I wonder if there is any downside of this solution.

PR #1549, opened 2017-01-24
chore(demo): accordion classname demo

musienkoyuriy:


demo/src/app/components/accordion/demos/styling/styling.css
```diff
@@ -0,0 +1,29 @@
+.tooltip.customClass .tooltip-inner {
```
valorkin: @musienkoyuriy may be remove tooltip css?

demo/src/app/components/accordion/demos/styling/styling.scss
```diff
@@ -0,0 +1,29 @@
+.tooltip.customClass .tooltip-inner {
```
valorkin: and completely remove scss

demo/src/app/components/accordion/demos/styling/styling.ts
```diff
@@ -1,8 +1,10 @@
-import { Component } from '@angular/core';
+import { Component, ViewEncapsulation } from '@angular/core';
 
 @Component({
   selector: 'demo-accordion-styling',
-  templateUrl: './styling.html'
+  templateUrl: './styling.html',
+  styleUrls: ['./styling.css'],
+  encapsulation: ViewEncapsulation.None
```
valorkin: shouldn't global styling work without VE.None?

valorkin (changes requested):

PR #1556, opened 2017-01-25
Fix rollup error.

jimmytheneutrino:
I couldn't use ng2-bootstrap's datepicker as is with rollupjs because rollup cannot process `import * as moment from 'moment'`, instead `import moment from 'moment'` should work.
See https://github.com/rollup/rollup/issues/670

### DIFF
--- a/src/datepicker/date-formatter.ts
+++ b/src/datepicker/date-formatter.ts
@@ -1,4 +1,4 @@
-import * as moment from 'moment';
+import moment from 'moment';
 
 export class DateFormatter {
   public format(date:Date, format:string):string {
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -13,6 +13,7 @@
     "declaration": true,
     "skipLibCheck": false,
     "stripInternal": true,
+    "allowSyntheticDefaultImports": true,
     "noUnusedLocals": false,
     "noUnusedParameters": false,
     "lib": ["dom", "es6"],


jimmytheneutrino:
Note that the travis error does not seem to be caused by the commit.

valorkin:
Thanks, I will try this with webpack and system.js and merge

jimmytheneutrino:
I realized that I actually did break typescript compilation. Setting
`"allowSyntheticDefaultImports": true`
fixes it. I am not sure if you want to go ahead with this option but IMHO it is not dangerous and is the simplest fix for the problem (one other option is, e.g., fixing moment's modules).

valorkin:
There are were an issue with such imports earlier, when ts typings was in issue

valorkin:
with `import moment from 'moment';`

I am getting
```
Error at /home/valorkin/work/open-source/ng2-bootstrap/src/datepicker/date-formatter.ts:1:8: 
Module '"/home/valorkin/work/open-source/ng2-bootstrap/node_modules/moment/moment"' has no default exp
```
from `ngc -p src` ;(
seems to be a dead end, we need to find a way how to configure rollup

jimmytheneutrino:
Yeah. See my previous comment. Setting "allowSyntheticDefaultImports" to true fixes it.

jimmytheneutrino:
Btw, if anyone needs this fix urgently, I temporarily published version 1.3.0 with this fix included at https://github.com/jimmytheneutrino/ng2-bootstrap/releases/tag/v1.3.0-fix.2 .

This should work with npm/yarn like this:

"ng2-bootstrap": "https://github.com/jimmytheneutrino/ng2-bootstrap/releases/download/v1.3.0-fix.2/dist.tgz"

miguelchico:
This error is happening again after:

https://github.com/valor-software/ngx-bootstrap/commit/1a516a611e091f09c76a233df0e4b07729bbf5b5

codez:
In the 1.7.1 npm, `date-formatter.js` still contains `import * as moment from 'moment';`, which fails the rollup process. Is this a regression or how does this statement get there?

valorkin:
Merge conflict, actually

jimmytheneutrino:
In our project we ended up creating a utility module a la

```
import * as realMoment from 'moment';
export const moment = (realMoment as any).default ? (realMoment as any).default : realMoment;
```

and using it like 
`import {moment} from './util/moment';` .

This should be ok for all environments.

PR #1574, opened 2017-01-29
Fix "Token must be defined!" AOT runtime error for PaginationComponent

lzoubek:
Fixes [PaginatonComponent issue](https://github.com/valor-software/ng2-bootstrap/issues/1444#issuecomment-275923716)


valorkin:
Thanks

PR #1603, opened 2017-02-05
docs(tabs): add to Tabs example access to static tabset from angular2 component

rafalkasa:
I have created this PR because I spent few hours, after I found out how to access from angular2 component to static tabset. I believe that this documentation will help others if they will have similar issue as I had.


valorkin (approved):
Looks good to me

PR #1604, opened 2017-02-05
comp(buttons) Fixed disabled buttons module tests

Le0Michine:
added fakeAsync and tick calls to force change detection work in tests


valorkin (approved):

valorkin:
Awesome

PR #1605, opened 2017-02-05
comp(dropdown) fixed disabled tests, removed outdated

Le0Michine:
fixed tests, removed outdated ones.
DropdownService made injectable and added to providers


valorkin (approved):
Great job as always!

Issue #1636, opened 2017-02-13
Permission denied when trying to push changes to branch


imribarr-compit:
Hey, I cloned the repository using SSH, created a branch from development branch and for some reason when I push the code to the branch I created, I get:
ERROR: Permission to valor-software/ng2-bootstrap.git denied to imribarr-compit.
fatal: Could not read from remote repository.

Also, I need to mention that committing the code only worked when I bypassed the commit hook.

Thanks.

SergeyKuryatnick:
@imribarr-compit Have you tried to use fork?

valorkin:
Thanks, I will try to merge it soon

PR #1639, opened 2017-02-13
- added typeaheadOnBlur event

imribarr-compit:
- added documentation for typeaheadOnBlur
- removed tab event key since blur event handles it
- added getter to _active item

### DIFF
--- a/src/typeahead/typeahead.directive.ts
+++ b/src/typeahead/typeahead.directive.ts
@@ -377,4 +377,4 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
   protected hasMatches(): boolean {
     return this._matches.length > 0;
   }
-}v
+}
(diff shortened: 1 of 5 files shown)


SergeyKuryatnick:
closes #1636

src/typeahead/typeahead.directive.ts
```diff
@@ -159,13 +162,6 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
       e.preventDefault();
       return;
     }
-
-    // if tab default browser behavior will select next input field, and
-    // therefore we should close the items list
-    if (e.keyCode === 9) {
```
valorkin, line 165: all looks good except of this change, it should not be a part of adding onBlur output

valorkin (changes requested):

valorkin:
@imribarr-compit looks good to me, but small change needed

valorkin:
BTW if you can add small test for this, it would be just awesome

src/typeahead/typeahead.directive.ts
imribarr-compit: I removed it since blur event handles it already, with this line, my change doesn't work.

imribarr-compit: Also, the only thing this line does is hides the results. This happens when the user blurs the component as well.

imribarr-compit: @valorkin

imribarr-compit:
@valorkin I've added a test for this. If you could just reply on my change request that would be great.

Thanks.

valorkin (approved):

valorkin:
Looks good to me!

PR #1651, opened 2017-02-15
Use [ngClass] to avoid conflicts with [class.x] bindings

### LINKED ISSUE #1650: Tabs - customClass bug
When using the [customClass] input, changes to it interfere with the active/disabled classes. This is due to binding both to class (through {{}} bindings) and [class.x]

This happens:

tabsbug

tiagoroldao:
Fixes #1650

### DIFF
--- a/src/tabs/tabset.component.ts
+++ b/src/tabs/tabset.component.ts
@@ -8,7 +8,7 @@ import { TabsetConfig } from './tabset.config';
   selector: 'tabset',
   template: `
     <ul class="nav" [ngClass]="classMap" (click)="$event.preventDefault()">
-        <li *ngFor="let tabz of tabs" class="nav-item {{tabz.customClass}}"
+        <li *ngFor="let tabz of tabs" [ngClass]="'nav-item ' + tabz.customClass"
           [class.active]="tabz.active" [class.disabled]="tabz.disabled">
           <a href="javascript:void(0);" class="nav-link"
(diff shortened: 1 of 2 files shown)


valorkin:
simple and great, can you add simple test to cover #1650 issue?

src/tabs/tabset.component.ts
```diff
@@ -8,7 +8,7 @@ import { TabsetConfig } from './tabset.config';
   selector: 'tabset',
   template: `
     <ul class="nav" [ngClass]="classMap" (click)="$event.preventDefault()">
-        <li *ngFor="let tabz of tabs" class="nav-item {{tabz.customClass}}"
+        <li *ngFor="let tabz of tabs" [ngClass]="'nav-item ' + tabz.customClass"
```
Le0Michine: it can be:
`[ngClass]="['nav-item', tabz.customClass]"`

just to avoid concatenation in the template

tiagoroldao: True, and I forgot to handle undefined values as well.

tiagoroldao:
@valorkin  tests added 👍 

Changed `expectActiveTabs` to use a `getTabItems` function as to break on my tests before this fix (rightfully), and because the style changes are due to `.nav-item.active`, not `.nav-link.active`.

valorkin (approved):
Awesome, thanks

PR #1676, opened 2017-02-22
fix(datepicker): update activeDate on select

fmalcher:
I wanted to use angular reactive forms and update the form value from within the component.
However, the date picker does not update its view then. This PR fixes the problem.

I saw there's a new datepicker as work in progress – great! 🎉 
Would be nice if you could merge this small fix for the meantime!

Best, Ferdinand


valorkin:
Thanks ;)

PR #1699, opened 2017-03-01
Update for typeahead.directive.ts

savinn:
When set to typeaheadMinLength ===0(instead ==0), typeahead doesn't show on focus the full list of options


valorkin:
let me guess, in your html it looks like:
`typeaheadMinLength="0"`?
if so try 
`[typeaheadMinLength]="0"`

savinn:
Thnx.It worked! But it is not in the documentation

valorkin:
Thanks for PR :)

PR #1700, opened 2017-03-01
WIP sortable

valorkin:
What I want to achieve:
- simple usage
- small and simple components
- under 5 inputs + under 5 outputs
What is missing:
- BsSortableConfig

I think users should be able to
```html
<div bsDropZone></div>
<div bsDropZone>
  <div bsDragable *ngFor="let item of items; let i = index;"><span>{{index}}></div>
</div>
```

should be enough to start using a sortable feature

the other thing, that there are could be `bs-sortable-list` and `bs-sortable-tree` components...

@Le0Michine what do you think?


valorkin:
+ plus bigger part of logic could be moved to injectable config\stateless service

Le0Michine:
@valorkin Actually for easy start there is a `Sortable` component it can render string items or custom template out of box. Regarding directives, in current implementation they rely on items list which is passed into `*ngFor`, they don't reorder dom elements but only items in passed collection so `*ngFor` does all the rendering stuff, unfortunately now I don't see how to avoid passing items list into the `DropZone` directive

Le0Michine:
@valorkin agree about logic, it needs some refactoring

PR #1703, opened 2017-03-02
feat(calendar): emit an event when View is changed

liviuciulinaru:
I have added a new event to know when the view of the calendar is changed (eg. user has changed the month). I am building a complex booking system and I need to calculate the disabled dates based on the current view (eg. when the user changes the month, a request to the server is made to get the disable dates). 

The event object will contain the activeDate and the datepickerMode.


valorkin:
Thanks!

PR #1706, opened 2017-03-03
fix(styling-local): fix tooltip arrow styles

Ugmaxie:
It has to be `border-bottom-color` to be able to change color of arrow.


valorkin:
@Ugmaxie have you checked both with bs3 and bs4 versions?
`npm run demo.serve` to check it locally

valorkin:
merged via https://github.com/valor-software/ng2-bootstrap/commit/b1f04d9af6ec5b02fb16331071641a114fd0422b

PR #1709, opened 2017-03-04
Correct Bootstrap name

trumbitta:
Bootstrap is not "Twitter Bootstrap" since like 2013 :)


valorkin:
Thanks, now I feel myself an old ;)

PR #1713, opened 2017-03-05
Instruction for Bootstrap 4 and @angular/cli

DouglasWebster:
I've written instructions on using Bootstrap 4 with ng2-bootstrap based on the  @angular/cli  instruction for including Bootstrap 4.  Hopefully this should provide a location for new user to discover the window.__theme = 'bs4' setting that seems cause a few problems.


DouglasWebster:
Wrong component chosen for the test - should have been a dropdown as collapse works without the window._theme being set.

PR #1718, opened 2017-03-07
feat(modal): make `show` config option work

### LINKED ISSUE #1680: ModalOptions.show property value is ignored
Based on documentation [here](http://valor-software.com/ng2-bootstrap/#/modals#modal-options), if `[options]="{ show: true }"` is provided, then the modal should be shown upon initalization. 
But this value is not even checked anywhere in [modal.component.ts](https://github.com/valor-software/ng2-bootstrap/blob/development/src/modal/modal.component.ts).

So, the option does not work. But it is a really nice one, since it allows to handle modals with `*ngIf` statements instead of `show() / hide()`

DethAriel:
Fixes #1680 . Adds code examples for utilizing this option. 

**IMPORTANT** this might have been a breaking change since the default value for "show" has changed. But since it looks like this option was never working in the first place - it should all be fine


valorkin:
fantastic, let me try it locally ;)

PR #1734, opened 2017-03-10
fix(dropdown,rating,typeahead): remove global in order to be usable …

karlhaas:
…with systemjs and aot


valorkin:
and here https://github.com/valor-software/ng2-bootstrap/pull/1734/files#diff-a4f2d9fe15a93898cef7b2fcb7adc4dbL45
try to replace with Event

and all other places

karlhaas:
why?

valorkin:
If not replace with Event it will throw an error at universal

valorkin:
Thanks

karlhaas:
@valorkin could you please provide a new version?

PR #1784, opened 2017-03-21
Include CDN instructions for Bootstrap 4

dgtlmoon:
Seems an easy way to get stuck for beginners, it's not clear that there's a version3, and version4 alternative


valorkin:
thank you

PR #1794, opened 2017-03-22
Update typeahead.directive.ts

Mohdplunk:
it is not seems to get integer type if set from the template


valorkin:
actually if `typeaheadMinLength="0"` from template it will be a string

valorkin:
I would rather stick to current strict behavior. thanks for PR!

PR #1812, opened 2017-03-25
Replace <template> with <ng-template>

yusijs:
`<template>` is deprecated as of angular@4.0.0 and needs to be changed to `<ng-template>`


yusijs:
The build fails due to requiring angular 4 - saw an open PR for it (https://github.com/valor-software/ng2-bootstrap/pull/1810). :)

valorkin:
Thanks, for PR but it is already available in @beta version :)
Better to add it to docs somewhere

PR #1814, opened 2017-03-25
Added minuteRoundBeforeStep option to timepicker

anandchakru:
if `minuteRoundBeforeStep` is `true` (default `true`), On first step, round current time to nearest minuteStep. 

**Example**: if current time is `05:23` and minuteStep is `5` increment will be in this order `05:25, 05:30, 05:35`


valorkin:
@anandchakru hey, I have completely reworked timepicker
you can see it here https://github.com/valor-software/ngx-bootstrap/pull/2058
hopefully you will like new version
please update PR to much new logic

PR #1832, opened 2017-03-29
Typeahead options added

joeoh1359:
Changes (after a bit of mess) are:
-Implemented typeaheadFocusFirst
-Added options for user to define Escape, Select, and to ignore keys (overriding defaults).


joeoh1359:
need to fix some bugs first

Issue #1848, opened 2017-03-31
Dropdown failed to run init.


hailg:
Hi, I'm using:

* @angular/cli: 1.0.0
* node: 7.5.0
* os: darwin x64

For now, I can't start using the dropdown component as it throwing the error as the image attached.
 
<img width="915" alt="screen shot 2017-03-31 at 6 19 28 pm" src="https://cloud.githubusercontent.com/assets/731508/24546665/ea8c8da4-163e-11e7-93d4-757d336b38d8.png">

This is my component html:

```html
<div class="jumbotron vertical-center" style="background-color: #fff;">
  <div class="container" style="text-align: center;">
      <div class="btn-group" dropdown>
        <button dropdownToggle type="button" class="btn btn-primary dropdown-toggle">
          Button dropdown <span class="caret"></span>
        </button>
        <ul *dropdownMenu class="dropdown-menu" role="menu">
          <li role="menuitem"><a class="dropdown-item" href="#">Action</a></li>
          <li role="menuitem"><a class="dropdown-item" href="#">Another action</a></li>
          <li role="menuitem"><a class="dropdown-item" href="#">Something else here</a></li>
          <li class="divider dropdown-divider"></li>
          <li role="menuitem"><a class="dropdown-item" href="#">Separated link</a>
          </li>
        </ul>
      </div>
  </div>
</div>
``` 

Is it something wrong with my code or this is a bug?

nielsvanlooy:
I'm having the same issue. There seem to have been some big changes to the dropdown last night, so i'm guessing this might be a bug

psaussure:
See #1838 for explaination: v1.6.0 introduced breaking change. You can see it in changelog. Solution: use 1.4.2 or any version prior breaking change. Alternate solution: change your code to replace module name according to changelog.

MarioGruda:
I have replaced the modulename and I iuse the templatesyntax '*dropdownMenu' but still have the problem.

`import { BsDropdownModule } from 'ng2-bootstrap/dropdown';`

`BsDropdownModule.forRoot(),`

```html <div class="btn-group pull-right" dropdown>
<button id="split-button" type="submit" class="btn btn-primary" [disabled]="searchForm.invalid">Suchen</button>
<button type="button" class="btn btn-primary dropdown-toggle-split" [disabled]="searchForm.invalid" dropdownToggle>
 <span class="caret"></span>
 <span class="sr-only">Suchen</span>
</button>
[...]

PR #1850, opened 2017-03-31
fix(dropdown): replace undefined filter-function (#1848)

krokerke:
The filter-function doesn't exist on EventEmitter.
instead we check the value inside the subscribe-function


valorkin:
or we can try to `import 'rxjs/add/operator/filter';`

valorkin:
like here https://github.com/valor-software/ng2-bootstrap/pull/1836/files

Thanks for PR!

PR #1891, opened 2017-04-11
Fixed example code for app.component.html

sumppi:
The previous example code resulted in broken widget, the fixed code is taken directly from http://valor-software.com/ngx-bootstrap/#/dropdowns#single-button


valorkin:
Thanks!

PR #1899, opened 2017-04-13
Change template to ng-template (Angular 4)

Valerika:
Angular desugars this notation into a marked-up <ng-template> that surrounds the host element and its descendants. Each structural directive does something different with that template.

https://angular.io/docs/ts/latest/guide/structural-directives.html

### DIFF
--- a/src/alert/alert.component.ts
+++ b/src/alert/alert.component.ts
@@ -5,17 +5,17 @@ import { OnChange } from '../utils/decorators';
 @Component({
   selector: 'alert,ngx-alert',
   template: `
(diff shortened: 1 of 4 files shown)


Valerika:
screenshot from 2017-04-13 19-40-28

ctrl-brk:
ng-template is Angular4 thing and I believe it's been already fixed in the upcoming version 2 (currently in beta)

valorkin:
Yes, it is already fixed. Seems it should be added to readmes.
Thanks for your efforts!

PR #1901, opened 2017-04-14
the previous example did not work properly

wdunn001:
the previous example dropdown was not formatted properly


valorkin:
Thanks! I will get back from vacation and will be merging it

valorkin:
Thanks!
fixed with #1891

PR #1905, opened 2017-04-16
fixing not working test example in /docs/getting-started/bootstrap4.md

ricevind:
Adding `dropdown-toggle` to `button` element for styling.
Adding `class="dropdown-menu"` to `ul` element to make example working.


valorkin:
Thanks!
fixed with #1891

Issue #1908, opened 2017-04-18
Add id parameter in TabDitective


akarabach:
I want create inline tabs like this 
```
<tabset>
  <my-wrapper>
    <tab [heading]="'tab1'" [id]="'unique1''">
      <tabset>
        <my-wrapper>
          <tab [heading]="'tab1'" [id]="'unique2'"></tab>
        </my-wrapper>
        <my-wrapper>
          <tab [heading]="'tab2'" [id]="'unique3'"></tab>
        </my-wrapper>
      </tabset>
    </tab>
  </my-wrapper>
</tabset>
```

Before i want control tabs from `my-wrapper ` (`@ContenChildren(TabDirective)`), but my tabs have the same heading and i don't how i can distinguish tabs by `heading`.
Please, add `id` parameter in` TabDirective`

PR #1909, opened 2017-04-18
Added id parameter in tab directive

akarabach:
For https://github.com/valor-software/ngx-bootstrap/issues/1908


valorkin:
Why do we need I'd param if it is not used?

akarabach:
@valorkin We integrated tabs with router and we control tabs from `<my-wrapper>` (see https://github.com/valor-software/ngx-bootstrap/issues/1908), but we have tabs with the same heading and we don't know how recognize tabs for to activate the desired tab.

valorkin:
Got your point, I will be back from vacation next week and will be merging PRs

PR #1916, opened 2017-04-19
fix dropdown missing outputs / emits on toggle, include & fix tests

### LINKED ISSUE #1915: Dropdown miss isOpen output, onShown / onHidden outputs emits only on body container
isOpen has no public output
onHidden / onShown emits only when ComponentLoader used ( container are set )

trimeyko:
Fixes #1915

added isOpenChange output to directive, added emits on show/hide for onShown/onHidden ( I'm not sure we should emit there something, in component loader componentRef.instance emitted btw.

moved spec file to spec/ folder & fixed tests, added few, removed tests related to dropdown-toggle class since it's not actual according to docs.

PS: currently there are no easy way to knew when dropdown is open/closed and no container are set for user, because no outputs provided, also that quite hard to close it programatically since isOpen on context is desynced with inner one only using various workarounds. Would be great if build will be available after we fix that.

### DIFF
--- a/src/dropdown/bs-dropdown.directive.ts
+++ b/src/dropdown/bs-dropdown.directive.ts
@@ -90,6 +90,10 @@ export class BsDropdownDirective implements OnInit, OnDestroy {
   }
 
   /**
+   * Emits an event when isOpen change
+   */
+  @Output() isOpenChange: EventEmitter<any>;
+  /**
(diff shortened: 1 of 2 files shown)


IlyaSurmay:
Thanks for PR! Outputs were added in #2006 and #1951

PR #1919, opened 2017-04-19
Fix(typeahead): show available choices onFocus when minLength == 0

### LINKED ISSUE #1559: Typeahead does not show list on focus when typeaheadMinLength="0" set
According to documentation this feature is already supported (when typeaheadMinLength="0", the dropdown should open on click listing available options).

But it does not work for me with version 1.3.1, nor it does on the demo page:
http://valor-software.com/ng2-bootstrap/#/typeahead#forms

(It work if after getting inside the input the backspace is pressed)

tzielins:
Fixes #1559 

Currently typeahead does not show list on focus when typeaheadMinLength="0" even is such behaviour is documented.

onFocus method is comparing minLength with 0, while the angular sets input property to string '0' and no event is emitted.

Fixed the issue by replacing === with == in the comparison.

### DIFF
--- a/src/spec/typeahead.directive.spec.ts
+++ b/src/spec/typeahead.directive.spec.ts
@@ -186,5 +186,58 @@ describe('Directive: Typeahead', () => {
       fixture.detectChanges();
     });
   });
+  
+  describe('onFocus', () => {
+    beforeEach(fakeAsync(() => {
+ 
+    }));
+
+    it('should result in a total of 2 matches, when minLength === 0', fakeAsync(() => {
+	  expect(directive).toBeDefined();
+	  
+	  directive.typeaheadMinLength = 0;
+      fixture.detectChanges();
+      tick(100);
+	  
+	  expect(directive.typeaheadMinLength).toBe(0);
+	  directive.onFocus();
+	  
+	  tick(100);
+	  
+	  expect(directive.matches.length).toBe(2);
+    }));	
+	
+    it('should result in a total of 2 matches, when minLength === \"0\"', fakeAsync(() => {
+	  expect(directive).toBeDefined();
+	  
+	  directive.typeaheadMinLength = '0';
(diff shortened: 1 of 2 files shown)


arun-haridas:
Guys , this issue is fixed or what??

renaud-dev:
Hello,
For people looking for a temporary fix, you can add square brackets arroud `typeaheadMinLength` : 

```
<form [formGroup]="myForm">
  <input formControlName="state"
         [typeahead]="states"
         [typeaheadMinLength]="0">
</form>
```
(angular 4.0.3 & ngx-bootstrap 1.6.6)

EDIT : contrary to what @dasdecoder pretends below, it works (confirmed by @tzielins). It makes `typeaheadMinLength` a basic regular Angular Input (taking advantage of template compilation, typing, etc...).

arun-haridas:
@renaudaste  I tried  `[typeaheadMinLength]="0"`  instead of `typeaheadMinLength="0"` , but still i'am having  the same issue ...!! Please help me to find a workaround.

tzielins:
@dasdecoder the [] trick works but you can also force the value to be 0 by declaring a numerical variable in your component with 0 value and reference it from the template, for example  [typeaheadMinLength]="zero"

valorkin:
Guys, it is not a trick. It works as expected
rather documentation should be updated
`[typeaheadMinLength]="0"` is completely correct behavior
`typeaheadMinLength="0"` is setting a string to numeric field,
it should not happen

Thanks for PR

arun-haridas:
@valorkin , @tzielins , @renaudaste  unfortunately none of the above mentioned solutions worked  for me.  I added @HostListener('click') listener in TypeaheadDirective.ts and  it started responding the way i need . I am not sure that it's the correct  way of solving the issue. Please correct me if am doing wrong...

PR #1955, opened 2017-05-04
Datepicker: use [hidden] instead of *ngIf for switching modes

timaxapa:
Hi. I use datepicker component in my app as a popup element. I created wrapper for datepicker with listener on outside clicks through ElementRef object like this:
`onDocumentClick(event) {
    if (!this.elementRef.nativeElement.contains(event.target)) {
      this.hideDatePicker();
    }
}`
but when I try to switch to the month view mode datepicker becomes hidden, because *ngIf removed my target component

So I suggest use [hidden] attribute binding for switching modes to get opportunity for inplementing popup mechanism: [VIDEO](https://drive.google.com/file/d/0B9EIP0vrFxAueUU4bHBQc2dKOWc/view?usp=sharing)


sivasunken:
Which version in npm is this fix available in?

valorkin:
new datepicker doesn't have such issue
available in v1.9.2
http://valor-software.com/ngx-bootstrap/#/datepicker

PR #1982, opened 2017-05-16
Atualização da diretiva do angular 4 ng-template

GustavoAmerico:


Martin-Luft:
Thanks for your PR.

For Angular 4 just use the @next version which fixes this warning. This cannot be fixed in 1.6.X because it would break Angular 2 support. You can use "ngx-bootstrap": "2.0.0-beta.2" or npm install ngx-bootstrap@next

Issue #1691, opened 2017-02-27
 Display multiple modals


kotproger:
For my project I need to open multiple modal windows. Using MODALS component it is possible to do, but when I close the top-level window, all previously open modal windows scroll bar disappears. Is there a solution to this problem?

kotproger:
For example, here is the code for displaying two pop-up windows that simulates the required behavior:
template
```
<button type="button" class="btn btn-primary" (click)="showExample()">Example</button>
<div bsModal #childModal1="bs-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="mySmallModalLabel" aria-hidden="true">
  <div class="modal-dialog modal-sm">
    <div class="modal-content">
      <div class="modal-header">
        <h4 class="modal-title pull-left">Child modal</h4>
        <button type="button" class="close pull-right" aria-label="Close" (click)="hideChildModal(1)">
          <span aria-hidden="true">&times;</span>
        </button>
      </div>
      <div class="modal-body" style="height: 1300px;">
        More Content Windo
      </div>
    </div>
  </div>
</div>

<div bsModal #childModal2="bs-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="mySmallModalLabel" aria-hidden="true">
  <div class="modal-dialog modal-sm">
    <div class="modal-content">
      <div class="modal-header">
        <h4 class="modal-title pull-left">Child modal</h4>
        <button type="button" class="close pull-right" aria-label="Close" (click)="hideChildModal(2)">
          <span aria-hidden="true">&times;</span>
        </button>
      </div>
      <div class="modal-body">
        Second Window
      </div>
    </div>
  </div>
</div>
```

component
```
import { Component, ViewChild } from '@angular/core';
import { ModalDirective } from 'ng2-bootstrap/modal';
 
@Component({
  selector: 'demo-modal-child',
  templateUrl: './child.html'
})
export class DemoModalChildComponent {
  @ViewChild('childModal1') public childModal1:ModalDirective;
  @ViewChild('childModal2') public childModal2:ModalDirective;
 
  public showExample():void {
    this.childModal1.show();
	this.childModal2.show();
  }
 
  public hideChildModal(n):void {
    this['childModal'+n].hide();
  }
}
```
The problem was solved, comment out the following lines in the file modal.component.ts (236)
```
  protected hideModal(): void {
    this._renderer.setElementAttribute(this._element.nativeElement, 'aria-hidden', 'true');
[...]

Issue #896, opened 2016-08-26
feat(modals): add support for nested modals (Modal removes the scrollbar to add its own)


Scipionh:
When I do open a modal on a page which has a scrollbar, the page is shifted to the right, as the page scrollbar is removed, then shifted back to the left (if the modal has its own scrollbar) which makes the whole thing very wonky.
It looks like opening a modal set overflow: hidden on the body tag (.modal-open class), and then add its own overflow-y: visible.
The same thing happens if you do close the modal.

I am yet to find a way to avoid this wonkyness. While the whole behavior makes sense, I am not sure how to prevent seeing the scrollbar switch.

PRR24:
Connected to the issue is if you have a modal that has a scroll and you open another modal on top of it, then after closing the topmost one, scrollbar for the first modal does not return and you can no longer scroll it (not tested in 1.1, is present in 1.0.24).
My typical use case of double modals:
... => edit item => modal dialog => changes to the item => are-you-sure modal => action => ...

valorkin:
nested modal, this use case not covered in twitter bootstrap, so this behavior simply not supported yet :)

nampdn:
@PRR24 I meet same problem like you, can't scroll the parent when the child modal is closed. Hope this behavior will support soon :)

Martin-Luft:
The first issue (page has scrollbars which disappear when opening a model) is  very annoying :(

Workaround:

`<body style="overflow-y: visible !important;">`

Martin-Wegner:
Workaround:

`<body style="overflow-y: visible !important;">`

szh:
Any update on this? I have Modal A which opens Modal B, but Modal B is bigger than Modal A, so it only shows up as big as Modal A, and has a vertical scrollbar. I think it should instead be appended to the body outside of Modal A.
Thanks

Martin-Luft:
Better workaround is to use in the main CSS:

```
.modal-open {
  overflow-y: auto;
}
```

IlyaSurmay:
Implemented by #1987

PR #1987, opened 2017-05-19
feat(modals): add support for nested modals (fix scroll)

H--o-l:
Twitter bootstrap (and so ngx-boostrap) doesn't officialy support nested modals.
You can however display nested modal but at least one issue appears:
At first modal you close, scroll events and scrollbar go back to the page, so
you can't scroll modals that are still displayed.

I propose to improve it with following modifications in modal component:
- at show, check if 'modal open' class is already attached to document body and
  set `isNested` boolean accordingly
- at hide, if nested, do not remove 'modal open' class and do not reset
  scrollbar.

Related to:
valor-software/ngx-bootstrap#896
valor-software/ngx-bootstrap#1691

### DIFF
--- a/demo/src/app/components/+modal/demos/index.ts
+++ b/demo/src/app/components/+modal/demos/index.ts
@@ -2,12 +2,15 @@ import { DemoModalSizesComponent } from './sizes/sizes';
 import { DemoModalStaticComponent } from './static/static';
 import { DemoModalChildComponent } from './child/child';
 import { DemoAutoShownModalComponent } from './auto-shown/auto-shown';
+import { DemoNestedDropdownsComponent } from '../../+dropdown/demos/nested-dropdowns/nested-dropdowns';
+import { DemoModalNestedComponent } from './nested/nested';
 
 export const DEMO_COMPONENTS = [
   DemoModalSizesComponent,
   DemoModalChildComponent,
   DemoModalStaticComponent,
-  DemoAutoShownModalComponent
+  DemoAutoShownModalComponent,
+  DemoModalNestedComponent
 ];
 
 export const DEMOS = {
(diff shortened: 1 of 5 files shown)


adrienverge:
Hi, what's the status on this?

It fixes the scroll problem on my project; I'd love to see it merged!

H--o-l:
@IlyaSurmay Thanks for the documentation !

YevheniiaMazur:
Сlicking on exit button on third modal window closes second modal window and makes third modal window inactive. 
Scroll was fixed

YevheniiaMazur:
Closing nested modals fixed. It looks good now

valorkin:
merged

Meligy:
I need to sit and give time to creating a proper bug report, but for now here's a quick note for anyone coming here via search like me:

This PR might have introduced a bug where calling `hide()` and `show()` multiple times synchronously, and some timing issue happens or something, causing the modal to think it's nested, and keeping the `ClassName.OPEN` class (`'modal-open'`) applied to the `<body>` tag even after modal is closed, which causes the page to lose its scrollbars because of `overflow:hidden;` style.

The current workaround I added to my application is adding another directive with the same selector as the modal component, that disables animation on the component itself so that all the logic is run synchronously. Something like:

```typescript
@Directive({
  selector: '[bsModal]'
})
export class BsModalOverrideDirective { 
  constructor(private modal: ModalComponent) { 
    modal.isAnimated = false;
  }
}
```

P.S.

Thanks heaps for all the great work on this awesome library.

valorkin:
@Meligy I hope to drop ng v2 support soon and move to ng v4.2
where we will have AnimationBuilder
so things should go better

valorkin:
BTW we are almost done with ModalService
https://github.com/valor-software/ngx-bootstrap/pull/2047

PR #2008, opened 2017-05-30
fix(popover): prevent ng router active link double ngOnInit issue

bruiztorres:
Same issue we had with the Dropdown directive. For some reason, the ngOnInit method is called twice.


valorkin (approved):

valorkin:
seems in latest router this issue is already fixed
but for those who use older versions it could be usefull
thanks for PR

bruiztorres:
;) @valorkin

bruiztorres:
I didn't have the time to add tests :(. 

@valorkin When are you expecting to release a new version?

valorkin:
There dedicated dev for ngx-bootstrap, plus I have some time lately. So releases will be weekly

PR #2010, opened 2017-05-30
fix(dropdown): fix duplicated events, add spec

IlyaSurmay:


YevheniiaMazur:
Duplicated events with container type "body" - weren't reproduce.
bs-dropdown.directive.spec - contains 4 Xited tests and unused variable addToggleClass

PR #2023, opened 2017-06-02
fix(1556): refix issue with moment import

sebek64:
- this was issue #1556, broken by 27a022922370947d54c3cfd357399eb588e8d01f

### DIFF
--- a/src/datepicker/date-formatter.ts
+++ b/src/datepicker/date-formatter.ts
@@ -1,4 +1,4 @@
-import * as moment from 'moment';
+import moment from 'moment';
 
 export class DateFormatter {
   public format(date:Date, format:string):string {


valorkin:
this is most changed line in this lib ;)
what works fine for one, doesn't work for another ;(

sebek64:
Hm, after some investigation, it seems that there is no good solution. The current state (1.7.0) is the way momentjs authors recommend. However, it is broken for Rollup users (or any ES6-module users in general). Is there any reason the import was changed back in the 1.7 version? I can't find it from the git changes, it is just part of a big commit.

valorkin:
@sebek64 I asked guys to add snowlake to ngx-bootstrap
so I could test build on rollup\gulp\system.js on each build

PR #2024, opened 2017-06-02
fix(accoridon): remove unneeded card-title class

7185:
This is only relevant with bs4.
You can see [here](http://valor-software.com/ngx-bootstrap/index-bs4.html#/accordion#simple) that the headers are bigger than their bs3 counterparts and that the titles aren't vertically centered.
This is because of the `card-title` class which adds a margin under those titles. This class is actually used wrongly here since it serves other purposes: adding a style on `<h*>` elements in the panel **itself** (not the header). See [the docs](http://v4-alpha.getbootstrap.com/components/card/#header-and-footer)  on Card Components.


valorkin:
Thanks for great explanation!

Issue #1795, opened 2017-03-22
Popover with triggers="focus" not working on MacOs


gilsdav:
On our projects and on your demo site, the popover with focus trigger doesn't work on Safari and Firefox (MacOsx). But it works on Chrome (MacOsx).

The original popover bootstrap works but not ng2-bootstrap popover.
[https://www.quirksmode.org/dom/events/blurfocus.html](https://www.quirksmode.org/dom/events/blurfocus.html)

> Firefox Mac and Safari and Chrome do not fire the events on radios and checkboxes if the user uses a mouse.

Within my test: On FF Mac, we can see the popover using tab key but not by clicking. On Safari, keyboard doesn't work.

A js fix to be able to handle focus event on button: http://jsfiddle.net/kzfg0pdh/6/

## Temporary fix
On component that directly use popover:
Use this on HTML
`triggers="focus:blur"`
On ts
`ngAfterViewInit() {
    const sels: any = document.querySelectorAll('[triggers]');
    for (let i = 0; i < sels.length; i++) {
        sels[i].addEventListener('click', function() {
            sels[i].focus();
        });
    }
}`

valorkin:
interesting

Kanishkaz:
Having the same issue. Firefox on windows doesn't like the focus trigger.

jboothe:
Same issue here on Mac Chrome, Firefox & Safari.

savyasachi-sa:
Facing the same issue. triggers=focus not working on Firefox Mac

liymike:
Having the same issue on mac with chrome firefox and safari, but works perfectly on windows using Microsoft Edge.
And also works on linux(opensuse) using firefox.

scndel:
The same issue for me with Firefox OSX & Safari on our project. Do you have any idea about potential fix or should we give up your solution ?

valorkin:
ok, it seems to be as designed (by browsers)
just add `tabindex="1"` to your element
and it works

gilsdav:
@valorkin 
Your solution 

> tabindex="1"
 
doesn't work for us. We already have the problem on Safari and Firefox.

Please reopen the issue (or accept my pull request).
Thanks ;)

scndel:
Hi, Gilsdav is damn right, that workaround is useless. You should merge his fix...

parasharabh:
Use <a> instead of <button> and add tabindex="0". This solved the issue for me.

ckrandhir:
[...]

PR #2031, opened 2017-06-03
Popover focus on Mac OS fix

gilsdav:
https://github.com/valor-software/ngx-bootstrap/issues/1795


gilsdav:
@valorkin  What's wrong ?

valorkin:
why popover needs focus?

gilsdav:
We need to hide it when clicking on other part of page. To be able to have blur event, we need focus before (like all Windows browser does but not some OSX browers). With this correction, it works like base bootstrap component (jquery).

SergeyKuryatnick:
Tested, looks good

valorkin:
Thanks for PR!

PR #2033, opened 2017-06-05
Remove dynamic tab with dynamic content and after add the same tab.

daibertdiego:
Fixing the problem when i remove a dynamic tab with dynamic content, and add this tab again to the view.

Ex:
`<tab *ngFor="let tabz of tabs"
                 [heading]="tabz.title"
                 [removable]="true">
                <div>
                    <app-custom-component></app-custom-component>
                </div>
            </tab>`

After remove some tab and try to add the tab to tabz array, the component doesn't appear.


valorkin:
native elements doesn't have `dispose` method

PR #2042, opened 2017-06-07
fix(positioning) don't modify readonly value

cmpilato:
(Positioning.position): element.getBoundingClientRect() returns a read-only value, so don't assign it directly to a variable whose properties are subsequently modified.


cmpilato:
So ... if I understand the `codecov` report correctly, because I took code that had no coverage and made it have more lines, I've made things worse.  Fair enough.

valorkin:
Thanks for PR, didn't know  that it is readonly!

PR #2057, opened 2017-06-13
fix(typeahead): Fix crash on Firefox and `contenteditable` input

adrienverge:
When using the typeahead directive on a `contenteditable` input, crashes can happen on Firefox:

    TypeError
    n.target.innerText is undefined

According to many sources \[1\] \[2\] \[3\], `innerText` should only be used on old browsers where `textContent` in not defined (e.g. Internet Explorer). On newer browsers that abide by standards, `textContent` is the right property to use to avoid problems.

\[1\]: https://stackoverflow.com/a/22990890
\[2\]: https://teamtreehouse.com/community/firefox-innertext-undefined-teachers-note
\[3\]: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

### DIFF
--- a/src/typeahead/typeahead.directive.ts
+++ b/src/typeahead/typeahead.directive.ts
@@ -118,11 +118,15 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
     }
 
     // For `<input>`s, use the `value` property. For others that don't have a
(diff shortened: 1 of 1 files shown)


adrienverge:
@valorkin I rebased on latest `development`!

valorkin:
Thanks for PR!

PR #2131, opened 2017-06-26
feat(modal): add dissmissReason, fix body padding, add events section…

### LINKED ISSUE #1964: feat(modal): provide an Output event when closing by Esc key
A modal template within a custom component might very well sport its own Ok and Cancel buttons. The former could invoke `this.modal.hide()` and then emit an event to client subscribers, like `this.myActionOk.next(...)`. The latter button could still invoke `this.modal.hide()` but then emit a `this.myActionCancel.next()`.

When modal has enabled closing by Esc key, and user presses that, there's no straightforward way for my wrapping component to notify subscribers about the fact that action has been canceled. And anyway `modal.hide()` is invoked behind the scenes, so modal is visually gone.

Even if wrapping component registers to `this.modal.onHide` output, that would be invoked whether modal is being closed for Ok button, or for Cancel button, or for Esc pressed.

Maybe an `onEscPressed` or, better, `onEscClosing` output event could fit the bill.

Any thoughts? Thanks!

### LINKED ISSUE #2004: ngx-bootstrap on modal implementaiton with angular 2.
Cause flickering in page - dynamic add padding-right to 15px;

IlyaSurmay:
closes #1964 
fixes #2004

### DIFF
--- a/src/modal/modal.component.ts
+++ b/src/modal/modal.component.ts
@@ -57,6 +57,8 @@ export class ModalDirective implements AfterViewInit, OnDestroy {
 
   // seems like an Options
   public isAnimated: boolean = true;
+  /** This field contains last dismiss reason. Possible values: `backdrop-click`, `esc` and `null` (if modal was closed by direct call of `.hide()`). */
+  public dismissReason: string;
 
   public get isShown(): boolean {
     return this._isShown;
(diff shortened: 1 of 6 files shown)


src/modal/modal.component.ts
```diff
@@ -89,7 +91,7 @@ export class ModalDirective implements AfterViewInit, OnDestroy {
     if (this.config.ignoreBackdropClick || this.config.backdrop === 'static' || event.target !== this._element.nativeElement) {
       return;
     }
-
+    this.dismissReason = 'backdrop-click';
```
valorkin: may be enum?

valorkin (approved):

PR #2190, opened 2017-07-14
feat(popover & tooltip): add container classes

### LINKED ISSUE #1395: Alternative to tooltip customClass deprecated
Hi, just updated my ng2bootstrap module and I am surprised that customClass is deprecated on tooltip...

Looking at the example provided, you have recently commented the part where customClass was applied. This is really important for me because we are showing a dt/dd inside the tooltip so we need it wider than 200px

There is some alternative?
Will it come back?

Thanks!

### LINKED ISSUE #1707: How can I override the style of a popover?
I have a popover in my component but it doesn't display correctly when I include the text on it. The text is overflowing the popover, I need the text not to be wrapped.

By doing my search on the internet, this is a common problem which gets solved by overriding the popover property "max-width" to be 100%, I proved that in the browser's debugger and it works. The problem is that I don't seem to be able to set/override this property from my component's css file.

I have tried:
`.popover {
  max-width: 100%;
}`

and

`.popover-content {
  width: 100%;
}`

and

`.popover-content, .popover-content .body {
  width: 100%;
}`

None of them have worked, any thoughts?

IlyaSurmay:
added ability to apply custom css classes to tooltip and popover
fixes #1707
fixes #1395

### DIFF
--- a/demo/src/app/components/+popover/demos/class/class.html
+++ b/demo/src/app/components/+popover/demos/class/class.html
@@ -1,5 +1,5 @@
 <button type="button" class="btn btn-primary"
-        popover="Vivamus sagittis lacus vel augue laoreet rutrum faucibus." [containerClass]="'customClass'">
+        popover="Vivamus sagittis lacus vel augue laoreet rutrum faucibus." containerClass="customClass">
   Custom class demo
 </button>
 
--- a/demo/src/app/components/+popover/demos/class/class.html
+++ b/demo/src/app/components/+popover/demos/class/class.html
@@ -1,0 +1,5 @@
+<button type="button" class="btn btn-primary"
+        popover="Vivamus sagittis lacus vel augue laoreet rutrum faucibus." [containerClass]="'customClass'">
+  Custom class demo
+</button>
+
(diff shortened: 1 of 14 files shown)


YevheniiaMazur:
Tested, looks good

demo/src/app/components/+popover/demos/class/class.html
```diff
@@ -0,0 +1,5 @@
+<button type="button" class="btn btn-primary"
+        popover="Vivamus sagittis lacus vel augue laoreet rutrum faucibus." [containerClass]="'customClass'">
```
valorkin: don't use `[]` binding for string setters
`[containerClass]="'customClass'"`
should be
`containerClass="customClass"`

demo/src/app/components/+popover/demos/index.ts
```diff
@@ -68,5 +70,9 @@ export const DEMOS = {
   stylingGlobal: {
     component: require('!!raw-loader?lang=typescript!./styling-global/styling-global.ts'),
     html: require('!!raw-loader?lang=markup!./styling-global/styling-global.html')
+  },
+  class: {
```
valorkin: using reserved keywords is no good

demo/src/app/components/+tooltip/demos/class/class.html
```diff
@@ -0,0 +1,5 @@
+<button type="button" class="btn btn-primary"
+        tooltip="Vivamus sagittis lacus vel augue laoreet rutrum faucibus." [containerClass]="'customClass'">
```
valorkin: string setter

valorkin (changes requested):

PR #2329, opened 2017-08-01
refactor(): deprecaded <template> element

danielsogl:
Changed deprecated `<template>` to `<ng-template>`


Martin-Luft:
Thank you for your PR.

For Angular 4 just use the @next version which fixes this warning. This cannot be fixed in 1.X.X because it would break Angular 2 support. You can use "ngx-bootstrap": "2.0.0-beta.2" or npm install ngx-bootstrap@next

PR #2352, opened 2017-08-03
Make the typeahead results show when typeaheadMinLength is 0 and the search string is empty

ddeschepper:
The != null comparison makes sure no null or undefined values are being emitted, while allowing for the empty string when the typeaheadMinLength is 0. Current behavior is to show the complete list on focus when typeaheadMinLength is 0, but when clearing the input value, it does not show the list.


YevheniiaMazur:
Tested, looks good

PR #2388, opened 2017-08-09
Fix carousel setInterval by running it outside of Angular zone.

aecz:
By doing so, Angular zone gets stable and Universal stops waiting indefinitely.
Also ApplicationRef.isStable works now as expected.


valorkin:
awesome! Thanks

YevheniiaMazur:
Tested, looks good

PR #2390, opened 2017-08-10
feat(typeahead): add dropup option

### LINKED ISSUE #2281: Typeahead option to pop results above input
I am using the typeahead at the bottom of a page, and I would like the option to have results pop above the input, instead of down as they go off the page.  Is this available?

IlyaSurmay:
closes #2281


YevheniiaMazur:
Tested, looks good

PR #2419, opened 2017-08-15
fix(dropdown): multiple dropdown perf issue

### LINKED ISSUE #2172: fix(dropdown): performance problems with multiple dropdowns
**Description**

**Workaround**

**Notes**

harunurhan:
* Fixes #2172

### DIFF
--- a/src/dropdown/bs-dropdown-toggle.directive.ts
+++ b/src/dropdown/bs-dropdown-toggle.directive.ts
@@ -1,5 +1,5 @@
 import {
-  Directive, ElementRef, HostBinding, HostListener, OnDestroy
+  Directive, ChangeDetectorRef, ElementRef, HostBinding, HostListener, OnDestroy, Renderer
 } from '@angular/core';
 import { Subscription } from 'rxjs/Subscription';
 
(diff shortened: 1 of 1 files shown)


harunurhan:
FYI: `npm run test` works and all tests pass locally.

valorkin (approved):

valorkin:
merged as part of https://github.com/valor-software/ngx-bootstrap/pull/2441
solves perf issue for all dynamic components

DLTKPeterEnevoldsen:
I don't think this solves the mentioned performance problem for dropdowns because there is still an onDocumentClick host binding on the dropdown-toggle directive. This listener will be always be attached and when we have many (1000s) of dropdown-toggles the performance will still be slow.

valorkin:
Yeah, we could not remove it, because of inline mode.
But there are one PR to remove this listener

DLTKPeterEnevoldsen:
Thank you for the quick reply. Will it be possible to include the PR in the next release.

valorkin:
Not sure, dev who was working on it, is on vacation this week

DLTKPeterEnevoldsen:
No problem. Thank you very much for the update.

harunurhan:
@PeterEnevoldsen well, this reduces number of global click listeners for dropdowns to a single one, which the issue mentioned, that's why I linked the issue but didn't look at it closely, sorry about that! 😟

PR #2428, opened 2017-08-16
feat(popover): support passing template context

### LINKED ISSUE #1682: Popover close from within the element itself
I want to create a popover with a template and within that template a button "close" wich closes this popover.
 I have the following code:

```
<button type="button" class="btn btn-primary" [popover]="options" placement="bottom" container="body">
    TemplateRef binding
</button>

<template #options>
    Just another template
    <div class="popover-footer">
        <button type="button" class="btn btn-default" (click)="options.hide()">Close</button>
    </div>
</template>
```

This gives me following error:

`self.parentView._TemplateRef_6_4.hide is not a function`

So I guess the id options isn't on the correct element.
Can someone please explain how I can make this work?

harunurhan:
* Fixes #1682 by adding a new @Input for popover directive so that
context can be passed when content is TemplateRef.

### DIFF
--- a/src/component-loader/component-loader.class.ts
+++ b/src/component-loader/component-loader.class.ts
@@ -107,13 +107,13 @@ export class ComponentLoader<T> {
   }
 
   // todo: appendChild to element or document.querySelector(this.container)
-  public show(opts: { content?: string | TemplateRef<any>, [key: string]: any } = {}): ComponentRef<T> {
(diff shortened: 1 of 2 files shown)


valorkin:
would be great to have a demo for this feature
thanks for PR

YevheniiaMazur:
Tested, looks good

PR #2442, opened 2017-08-17
Add `onDrop` Event to sortable

tandibar:
I have a collection with an order which should be synced with the server. The `onChange` event is fired continuously on dragging and does not fullfill my needs.

That is why i introduced the `onDrop` callback to be notified whenever an element is dropped.


tandibar:
I just realised that i can solve my problem by adding `(drop)="someCallback()"` to my sortable elements. So perhaps it makes no sense to add this functionality.

valorkin:
@tandibar thanks for PR, I suppose you are correct, so I will close this one

PR #2450, opened 2017-08-17
Use <ng-template> instead of <template>

paavieira:
The `<template>` element is deprecated. Use `<ng-template>` instead.


fchiumeo:
please, please search before creating new issues
https://github.com/valor-software/ngx-bootstrap/search?utf8=%E2%9C%93&q=ng-template&type=Issues

npm i ngx-bootstrap@next

PR #2465, opened 2017-08-21
feat(datepicker): removed dependency on moment.js

valorkin:


YevheniiaMazur:
Tested, looks good

src/datepicker/bs-datepicker.module.ts
```diff
@@ -34,6 +32,13 @@ import { BsDatepickerDayDecoratorComponent } from './themes/bs/bs-datepicker-day
     BsDatepickerComponent, BsDaterangepickerComponent]
 })
 export class BsDatepickerModule {
+  constructor() {
+    if (isDevMode()) {
+      console.warn(`BsDatepickerModule is under development,
+      BREAKING CHANGES are possible,
+      PLEASE, read changelog`);
+    }
+  }
```
FrancescoBorzi, line 41: I guess we should exclude this message from tests

image

valorkin: question is how to figure out that we are under testing?
what I can do is to show this warning only once (as was initially planned):
https://github.com/valor-software/ngx-bootstrap/pull/2585

Issue #2536, opened 2017-08-29
AlertComponent is still using template instead of ng-template. 


avarndell:
I just installed 1.9.1

devstu:
Just installed. Angular-CLI, followed the directions. Same issue.

ctrl-brk:
Use ngx-bootstrap@next

devstu:
That worked for me. Thank you.

avarndell:
Thanks.  That worked.

PR #2537, opened 2017-08-29
Update ng-cli.md

devstu:
https://github.com/valor-software/ngx-bootstrap/issues/2536


valorkin:
hey, thanks for PR,
but it rather should be a note
that for ng v2+ install ngx-bs
and for ng v4+ install ngx-bs@next

this week will be last release supporting ng v2
and next 2 weeks we will work on upgrading to ng v4

valorkin:
any way, @next will became @latest next week

PR #2540, opened 2017-08-30
feat(datepicker): added month view

valorkin:
added month\year selection views
added min\max dates boundaries


YevheniiaMazur:
Tested, looks good

PR #2580, opened 2017-09-04
feat(core): try to guess bs version

valorkin:
- _guessBsVersion should be called only once
- check demo.serve
- remove window.__theme from html for demo


karma.conf.js
```diff
@@ -15,6 +15,7 @@ module.exports = function (config) {
       require('@angular/cli/plugins/karma')
     ],
     files: [
+      {pattern: './node_modules/bootstrap/dist/css/bootstrap.min.css', included: true, watched: false},
```
valorkin: @IlyaSurmay does it works? or forgot to remove?

IlyaSurmay: It does. Without css result will be the same as it was before.

PR #2600, opened 2017-09-08
feat(modal): resolve modal data before OnInit

### LINKED ISSUE #2530: feat(modal): initialize and create all component variables before OnInit
When you create a popup like this (using newly implemented #2047 - modal service)

``` typescript
    this.bsModalRef = this.modalService.show(ModalContentComponent);
    this.bsModalRef.content.title = 'Modal with component';
```
It seems that it assigns variable (title) **after** the popup is created. If this popup contains other components itself, that take that variable as an input parameter, this may cause things to be broken in our case :(

Is it possible to initialize all variables in the component being created **before it is rendered first time**?
I  am probably asking exactly for the same feature we had in ng-boostrap (angular-js), something like "resolve" parameter for "show()" function that resolves all arguments when creating the dialog (set them after creating the component, but before adding it to the DOM / showing it)? Something like:

``` typescript
    this.bsModalRef = this.modalService.show(ModalContentComponent, {
       resolve: {
           title: () =>  'Modal with component'
        }
    });
```

### LINKED ISSUE #2733: Sending data to modal through service
I see in the examples that you can pass data to the model component through the service by doing something like `this._bsModalRef.content.name='test';` 

When I try to access this data however, I am always told that content is undefined. 

Is there a way to pass data and have it available in the `ngOnInit` method so that I can use data thats passed in my model logic?

Currently I have to use behavior subjects and have my model subscribe to them so I can pass it data to use which is a pain.

Am I missing something here?

IlyaSurmay:
Closes #2530
Closes #2733


nbelyh:
Thank you for the fix!
Is there any chance for this to be merged in the next release?

valorkin:
I would say chances are 100%

vanackere:
Hi,

 Any ETA on merging this pull request ? This particular feature is very useful and it would be nice to have it on a released version.

vanackere:
Hi,

 Any ETA on merging this pull request ? This is a really useful feature and it would be really nice to have this in a real release.

vanackere:
Hi again,

 I've been running a local fork of ngx-bootstrap with this pull request applied in production for a few weeks with no problem (the alternative way of using a global service to pass parameters to the modals was definitely too ugly).
 It looks like 2.0.0-rc.0 is out but still misses this change... is anything particular - like known bugs or regressions- holding this from being merged ?

valorkin:
My time, it will be merged prior to releasing v2

src/component-loader/component-loader.class.ts
```diff
@@ -117,28 +117,23 @@ export class ComponentLoader<T> {
...
-    opts: {
-      content?: string | TemplateRef<any>;
-      context?: any;
-      [key: string]: any;
-    } = {}
+
+  show(opts: {
+    content?: string | TemplateRef<any>;
+    context?: any;
+    data?: any; [key: string]: any;
```
valorkin: `initialState`

valorkin (changes requested):

valorkin (dismissed):

YevheniiaMazur:
Service modals do not work correctly - they are throwing error after opening and stop working (only reload page helps for closing)
errorsmodals

YevheniiaMazur:
tested, looks good

valorkin (approved):

PR #2602, opened 2017-09-08
feat(package): drop support of ng v2 and add support of ng v4 and v5

valorkin:


fchiumeo:
feat(package): drop support of ng v4 and add support of ng v5
you mean
feat(package): drop support of ng v2 and add support of ng v4 ?

valorkin:
@fchiumeo indeed :) thanks, seems it's enough for today

Issue #2617, opened 2017-09-11
Accordion: `isOpen` should be two-way bindable


steelsojka:
The accordion-group component should be able to be bound 2 ways, since the `isOpen` state can be modified from inside the component and from outside it. The problem scenario occurs when `isOpen` is bound from outside the component and the user clicks on the accordion. Internally the state is changes, but the binding has not. The change is not communicated back to the parent component.

Instead of the internal state being modified it should emit through an output named `isOpenChange` with the new value. This will sync the states.

This will be a breaking change however since, currently, if the user clicks on a group when `isOpen` is bound they still toggle the open state. This change would only respect the bound value. It would require a 2 way data bind to produce the expected behavior.

steelsojka:
The other alternative is to at least provide an output with the new state to be able to update the parent component. This would be the easiest and non breaking change.

CodisRedding:
Is two-way binding of isOpen for the accordion implemented in 3.3.0? I'm using isOpen in the component and it will not collapse a group except for the first time.

PR #2619, opened 2017-09-11
feat(accordion): add output for isOpen state changes

### LINKED ISSUE #556: accordion events
I want to do things when user switches [accordion-group](https://github.com/valor-software/ng2-bootstrap/blob/master/components/accordion/)s of accordion component. Say, I want to save in application state information about which accordion-group is currently open.

Maybe I missing something in the way angular2 does things, but I think this kind of behavior could be achieved with events. Some kind of event like _isOpenChanged_ for accordion-group seems missing to me.

How can I bind some code to user actions of opening and closing accordion groups? Am I right that events are missing?

steelsojka:
This fixes issue https://github.com/valor-software/ngx-bootstrap/issues/2617 without making any breaking changes. This can solve several issues with accordion open state as well.

Closes #556


IlyaSurmay:
@steelsojka hey, thanks for PR!
only one note, `isOpenChanges` should be renamed to `isOpenChange`
so you can use two way data binding

steelsojka:
@IlyaSurmay I thought about that, but I thought it would be a breaking change since it would have broken one way bindings. I was thinking you would need to emit on `isOpenChange` instead of modifying the internal state, but you could modify the state and notify without breaking current behavior. I'll make the change.

IlyaSurmay:
Tests are failing :( Please fix them, then we'll merge it.

valorkin (approved):

valorkin:
@steelsojka thanks for your contribution! ;)

psined1:
Hi, any idea when is this going live?

valorkin:
I am trying to publish small versions each Thursday, will be available as @next version
docs are here http://ngx-bootstrap.surge.sh/#/

steelsojka:
@IlyaSurmay Thanks for fixing these tests and getting this working. I've been busy and wasn't able to get to them. I look forward to seeing this in a release!

PR #2744, opened 2017-09-28
feat(datepicker): disable specified days of week

ahtshamraziq:
added `dayDisabled` prop that accepts an array of integers (0=Sunday, ..., 6=Saturday) and disables the corresponding days of the week.


ralzinov:
Anybody know is that PR would be merged?

valorkin (approved):

YevheniiaMazur:
tested, looks good

valorkin:
thanks a lot! :)

KartikShankarappa:
@valorkin , What if we want to disable specific dates in each month ?
For eg: I want to disable all the dates in all months and in all the years except 15th and 20th of each month.
Right now, the current version of datapicker gives us the `datesEnabled` and `datesDisabled` properties. But that would mean I will have to calculate such dates myself. Also it would force me to put a Min and Max date between which i will have to calculate the enabled and/or disabled dates.
How about a function that will help us to enable or disable specific dates ?
Something similar to [this](https://ng-bootstrap.github.io/#/components/datepicker/overview#limiting-dates)

PR #2776, opened 2017-10-02
refactor(gettign started): changed Getting started page structure

EvilAlexei:
- Replaced content from md file to html
- Removed old `public` attribute from demo components


valorkin:
@EvilAlexei is it ready for merge?

PR #2813, opened 2017-10-06
feat(tabs): add opportunity to add multiple classes to customClass attribute

### LINKED ISSUE #2467: Tab: customClass doesn allow multiply class
Hi, 
My code is .... [customClass]="getStyles()"
getStyles returns "hide" or "hide invalid". In case "hide invalid" I get error "The token provided ("hide invalid") contains HTML space characters, which are not valid in tokens." because of splitting classes by space.
This is because in beta 4, for customClass setter you use this.renderer.setElementClass which does not allow multiply classes separated by space.
How can i continues using multiply classes for customClass?

LimarenkoDenis:
Fixes #2467

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.

### DIFF
--- a/src/tabs/tab.directive.ts
+++ b/src/tabs/tab.directive.ts
@@ -29,21 +29,15 @@ export class TabDirective implements OnInit, OnDestroy {
   }
 
   set customClass(customClass: string) {
-    if (this._customClass && this._customClass !== customClass) {
-      this.renderer.removeClass(
-        this.elementRef.nativeElement,
-        this._customClass
-      );
-    }
-
     this._customClass = customClass;
+    this._customClass.split(' ').forEach((element: string) => this.setElementClass = element);
+  }
 
-    if (this._customClass) {
-      this.renderer.addClass(
-        this.elementRef.nativeElement,
-        this._customClass
-      );
-    }
+  set setElementClass(customClass: string) {
+    this.renderer.addClass(
+      this.elementRef.nativeElement,
+      customClass
+    );
   }
 
   /** tab active state toggle */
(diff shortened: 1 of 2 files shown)


IlyaSurmay:
Hi, thanks for PR!
We can't merge it because looks like it hasn't been tested. This is what I get on tab init stage

Please, fix your PR, test it and update, then we'll merge it.

src/tabs/tab.directive.ts
```diff
@@ -29,21 +29,15 @@ export class TabDirective implements OnInit, OnDestroy {
...
   set customClass(customClass: string) {
-    if (this._customClass && this._customClass !== customClass) {
-      this.renderer.removeClass(
-        this.elementRef.nativeElement,
-        this._customClass
-      );
-    }
-
     this._customClass = customClass;
+    this._customClass.split(' ').forEach((element: string) => this.setElementClass = element);
```
IlyaSurmay: this will throw an error if customClass isn't set

IlyaSurmay: this will throw an error if customClass isn't set

src/tabs/tab.directive.ts
```diff
@@ -29,21 +29,15 @@ export class TabDirective implements OnInit, OnDestroy {
...
+  }
 
-    if (this._customClass) {
-      this.renderer.addClass(
-        this.elementRef.nativeElement,
-        this._customClass
-      );
-    }
+  set setElementClass(customClass: string) {
+    this.renderer.addClass(
```
IlyaSurmay: what if list of classes changes? for example, if user pass `class-1 class-2` and then `class-2 class-3`. in this case all 3 classes will be appended

IlyaSurmay (changes requested):

valorkin:
Thanks for PR :)

PR #2816, opened 2017-10-09
fix(typeahead): fix close on blur

### LINKED ISSUE #2588: Typeahead within modal closes when scrollbar is clicked
Here's a plunker:
https://plnkr.co/edit/r0xqdcLOmGzfEYKc7H8r

Don't click "open modal" button just yet.
Type something (ex: letter a) in the input box, make your browser window small enough so you have a scrollbar on the "run" window. Play with scrollbar - everything works as expected.

Now open the modal window and try the same within it (there's another typeahead control). As soon as you touch the scrollbar control closes (receives blur event)

IlyaSurmay:
Fixes #2588 

 - [x] added/updated tests.

### DIFF
--- a/src/typeahead/typeahead.directive.ts
+++ b/src/typeahead/typeahead.directive.ts
@@ -131,11 +131,12 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
 
   private _typeahead: ComponentLoader<TypeaheadContainerComponent>;
   private _subscriptions: Subscription[] = [];
+  private _outsideClickListener: Function;
 
   constructor(private ngControl: NgControl,
               private element: ElementRef,
               viewContainerRef: ViewContainerRef,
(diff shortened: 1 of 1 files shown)


src/typeahead/typeahead.directive.ts
```diff
@@ -276,6 +276,10 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
         dropup: this.dropup
       });
 
+    this._outsideClickListener = this.renderer.listen('document', 'click', () => {
```
valorkin, line 279: host('document:click') ?

IlyaSurmay: It was done this way to achieve better performance. A lot of host listeners will decrease it so I decided to attach listener only when box with results is open.

PR #2918, opened 2017-10-26
feat (datepicker) - Set startingDay of the week via bs-config

### LINKED ISSUE #2622: Setup startingDay/first day of week to monday
Hello,

I would like to ask how to set first day of the week to Monday. As I read through the library I can see that in the DatepickerConfig there is option to set up startingDay to something else as 0 (which defaults to Sunday).
However setting like:

<input class="form-control" placeholder="Datepicker" bsDatepicker **[startingDay]="1"** **[bsConfig]="options"** [minDate]="minDate" [maxDate]="maxDate" #dp="bsDatepicker">

gets me an error:
Can't bind to 'startingDay' since it isn't a known property of 'input'

and bsConfig has no such property as starting day or something similar.
Changing local does not effect this option. What am I missing ? Big thanks

SidekickJohn:
 - [ ] added/updated tests.
 - [x] added/updated demos.

Closes #2622 
@valorkin @IlyaSurmay  please check this and tell me where I can help to finish this.

### DIFF
--- a/demo/src/app/components/+datepicker/datepicker-section.component.html
+++ b/demo/src/app/components/+datepicker/datepicker-section.component.html
@@ -16,6 +16,7 @@
             <li><a routerLink="." fragment="disabled">Disabled</a></li>
             <li><a routerLink="." fragment="forms">Forms</a></li>
             <li><a routerLink="." fragment="reactive">Reactive forms</a></li>
+            <li><a routerLink="." fragment="startingday">Set a starting Day for the week</a></li>            
           </ul>
         </li>
         <li><a routerLink="." fragment="api-reference">API Reference</a>
@@ -98,6 +99,13 @@
       <h3 routerLink="." fragment="reactive" id="reactive">Reactive forms</h3>
       <ng-sample-box [ts]="demos.reactive.component" [html]="demos.reactive.html">
         <demo-datepicker-reactive-forms></demo-datepicker-reactive-forms>
+      </ng-sample-box>
+
+      <h3 routerLink="." fragment="startingday" id="startingday">Setting a different start day of the week</h3>
+      <ng-sample-box [ts]="demos.startingday.component" [html]="demos.startingday.html">
+          <p>You can change the starting day of the week by changing <code>startingDay</code> property in <code>bsConfig</code> object. <br/>
+          Values 0 - 6 resemble Sunday - Saturday.</p>
+        <demo-date-picker-startingday></demo-date-picker-startingday>
       </ng-sample-box>
 
       <h2 routerLink="." fragment="api-reference" id="api-reference">API Reference</h2>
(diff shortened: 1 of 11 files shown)


PerfectlyNormal:
Nice work!

Would be slightly better if it read the start date from the configured locale. The locales already contain this, as `week.dow` (https://github.com/valor-software/ngx-bootstrap/blob/663f07899113f6b58c5164851fd2d345156ddd06/src/bs-moment/i18n/sv.ts#L58)

SidekickJohn:
@PerfectlyNormal thanks for the hint. <br/>
I got an other suggestion. <br/>
How about making this `week.dow` as default value only if there is no other value provided via bs-config?
That way you could still set a divergent start-day-of-the-week if you need it (which could be the case). <br/>
But before I'm adding this I will wait to see what the gatekeepers think of my existing solution.

PerfectlyNormal:
@sidekickjohn Yes, that is the ideal way. I guess I expressed myself poorly, as that was what I really want.

andre-c-andersen:
This looks good. Looking forward to it. Was about to look for another date picker, but since this looks to be coming in the not to far off future, I'll wait.

valorkin:
hey, I was absolutely sure that this issue is fixed, sorry
before merging this PR I want to fix: getting the first day of a week from locale
then this PR may not be needed

@SidekickJohn thank you so much for the contribution! :)

SidekickJohn:
@valorkin getting the first day of the week from locale would be the best thing to start with, so thank you! But don't you think setting a different start day via bs-config might be useful for some other use cases?

valorkin:
@SidekickJohn it could be, we will see

valorkin:
@SidekickJohn thanks again, please rebase your PR on top of dev
after merge of https://github.com/valor-software/ngx-bootstrap/pull/2970
now it should respect locale! :D

andre-c-andersen:
@SidekickJohn and @valorkin. Configuring it only using the local is indeed a first step, but remember that locale also changes the language of the date elements. This means a US English interface which uses the ISO 8601 date standard (i.e., Monday first) won't be possible. We're working on a Scandinavian webapp, thus language localization isn't important just yet (and might not ever be), but date localization is. Having Sunday as the first day of week is very foreign to our users and will make the app look sloppy. Having Monday as the first day of week and an arbitrary European language for the date element names is even worse.

valorkin:
There are en-gb locale in English and Monday first

andre-c-andersen:
@valorkin Thanks. That could work for us. Doesn't seem like there are any unfortunate language idiosyncrasies showing up either. So, that's good. That said, there are probably plenty of other use cases where a forced first weekday override makes sense.

valorkin:
@SidekickJohn thanks for PR again:)
please check our latest docs structure
ask @EvilAlexei if you will have any questions

valorkin:
with this merged: https://github.com/valor-software/ngx-bootstrap/pull/3271
you will be able to have more granular control over locales

valorkin:
now it's feasible via locale :)

wautersj:
I've read most of the tickets and comments and am not able to change the first day of the week. 

In our application the user can pick it's own day (one of the seven) as his preferred starting day of the week.
And I should configure the datePicker accordingly, of course.

When I look at the BsDatepickerConfig, there seem to be no option to do this. 
But I see people are confirming this is possible (!?)

Can someone explain to me, how to do this?

We're currently running: "ngx-bootstrap": "^5.6.2"



SO much thanks in advance!!

PR #2925, opened 2017-10-27
feat(demo): add universal support, add dockerfile

### LINKED ISSUE #2778: Tabset component removeTab method doesn't work with Angular Universal
- Angular version: 4.4.3
- ngx-bootstrap version: 1.9.3
- build system: angular cli (Universal + AoT)
- Use case: Tabset component doesn't work with Angular Universal

There's a bug in the tabset component that's causing an error with angular universal.

The error I'm getting is:
```
TypeError: tab.elementRef.nativeElement.parentNode.removeChild is not a function    
at TabsetComponent.removeTab ...
```
I think the issue is caused because you are accessing the DOM directly in this line:
https://github.com/valor-software/ngx-bootstrap/blob/b9ee70600e542ba2ea739976ed78af94aea19c2b/src/tabs/tabset.component.ts#L83

Maybe there's an easy fix for this using the [Renderer2 removeChild method](https://angular.io/api/core/Renderer2#removeChild) instead of `parentNode.removeChild`

I think this is a very important issue because a small issue is preventing ngx-bootstrap to work correctly with Angular Universal. After [searching in this repo](https://github.com/valor-software/ngx-bootstrap/search?utf8=%E2%9C%93&q=removeChild&type=), I found that this same issue may be affecting the modal directive as you use the `parentNode.removeChild` directly through the DOM instead of using the Renderer2 method.

IlyaSurmay:
Closes #2778


valorkin (dismissed):

YevheniiaMazur:
Warnings Unable to parse SourceMap appears and there are no dropdown with versions (both for demo.serve and demo.serve-universal run)
demouniversal

valorkin:
@YevheniiaMazur we don't care about source maps

PR #2930, opened 2017-10-28
Fix/2929 display date on init

### LINKED ISSUE #2929: datepicker don't display initial value in <input>
I init `bsValue` with current date, but datepicker input is empty. Datepicker store this value because when I toggle it I can see that current date is selected

screen shot 2017-10-28 at 21 05 50

https://stackblitz.com/edit/ngx-bootstrap-datepicker

ngx-bootstrap: 1.9.3

Angular: 4.3.6

Bootstrap: 3.3.7

grzegorz-jazurek:
- fix #2929 

I found out that Input interceptor for bsValue emit bsValueChange event before bs-datepicker-input.directive is initialised https://github.com/valor-software/ngx-bootstrap/blob/4e3e456251c64e0d279b731807667de66496a097/src/datepicker/bs-datepicker-input.directive.ts#L35

I used `setTimeout` to move execution at the end of stack. 
I'm new in angular 4 so maybe you can use better solution, but main problem is defined above

### DIFF
--- a/src/datepicker/bs-datepicker.component.ts
+++ b/src/datepicker/bs-datepicker.component.ts
@@ -69,6 +69,7 @@ export class BsDatepickerComponent implements OnInit, OnDestroy, OnChanges {
       return;
     }
     this._bsValue = value;
+    // wait for initialization bsValueChange subscribers
     setTimeout(() => {
       this.bsValueChange.emit(value);
     });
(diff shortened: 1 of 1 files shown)


valorkin:
@grzegorz-jazurek thanks a lot for issue and PR!
fixed in a bit different way: https://github.com/valor-software/ngx-bootstrap/pull/2962

PR #2942, opened 2017-11-01
fix(pagination): and pager to mark for check when page is changed

valorkin:


YevheniiaMazur:
Needs fix for Bs3. Pagination. Pager example. Default buttons appears under next and previous buttons when user click on them
screenshot from 2017-11-01 16-00-41

valorkin:
@YevheniiaMazur css issue, should be solved separatly

Issue #1217, opened 2016-11-08
dropdownToggle directive stops and prevents click events


milichev:
We need to catch click events on the document level, but `dropdownToggle` directive has hardcoded `stopPropagation` and `return false` (which leads to `preventDefault`) in its click handler. Can it be changed? Either remove both or move to `if` with a property like so:

```TypeScript
  @Input() public suppressClick:boolean = true;
  
  @HostListener('click', ['$event'])
  public toggleDropdown(event:MouseEvent):boolean {
    if (this.suppressClick) {
      event.stopPropagation();
    }

    if (!this.isDisabled) {
      this.dropdown.toggle();
    }

    return !this.suppressClick;
  }
```

valorkin:
Hm, this is how it is in twitter bootstrap. But we can make it configurable

H--o-l:
Same issue here, related to line:
https://github.com/valor-software/ngx-bootstrap/blob/development/src/dropdown/bs-dropdown-toggle.directive.ts#L44

I think the issue can be reopen :-)

valorkin:
@H--o-l need to remove it on one side, and have controversial requirements to keep it :(

H--o-l:
@valorkin thanks for the quick answer !
In the meantime I opened the PR #2964 on the subject. Let me know what you think about it and if it's possible to merged it.
If it's not possible I will find another way but I think the option doesn't hurt. You let me know !

Issue #2800, opened 2017-10-05
Multiple dropdowns (menus) stay open at the same time


ctrl-brk:
### Bug description or feature request:
I'm using dropdowns as a header menu. Since beta.5 when I click on the other menu the previous one stays open. I can have all my menus open at the same time. That's not happening in beta.4

### Plunker/StackBlitz that reproduces the issue:
It might be a side effect of nested dropdowns.
You can kind of see it here:
https://valor-software.com/ngx-bootstrap/#/dropdowns#nested
Open the dropdown (btw, it will not close when you click outside like others do), hover nested dropdown items and you'll be able to have both of them open.

### Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: 2.0.0-beta.6

Angular: 4.4.4

Bootstrap: 3

valorkin:
It was like that from the start, unfortunately, I didn't figured out how to make it ... better
But you can achieve better behavior with css

ctrl-brk:
As I said it doesn't happen in beta.4. Started in beta.5

ctrl-brk:
Here's the plunker. Click on top level menus - you will see the issue.
Change ngx-bootstrap version to 2.0.0-beta.4 and the issue disappears.
https://plnkr.co/edit/OEtnl6?p=preview

ctrl-brk:
Still happening in beta 7

valorkin:
True

ctrl-brk:
Now that's become a real problem. I can't use beta.4 with just released Angular 5 because of the ngOutletContext and I also can't use latest version because of this issue. Which pretty much means that I can't use Angular 5 with current ngx-bootstrap.

valorkin:
It's in the list to fix today, and to be in beta.8

valorkin:
@ctrl-brk found where an issue came and have a fix, just in case I have 2 envs now with ng 4 and 5.
So when we will publish, it should just work :)

valorkin:
Fix published ;)

chelaruc:
Hi guys! Thank you for your work ❤️ . Unfortunately, I still have this issue on v5.3.2. Can you help me?

manar-mk:
Hi, I'm facing the same problem when using `$event.stopPropagation()`. 
I know, it's not an issue, but maybe any ideas how to workaround?

aleksandar-pajic-44:
Any update on this?

PR #2964, opened 2017-11-02
feat(dropdown): add option to let event propagate on toggle

H--o-l:
Issue valor-software/ngx-bootstrap#1217 was fixed but re-introduce
recently in development branch (commit 04cab1eb6f42ad7a6456c34a0ec54f9bd37fd83a).

Let's add an option to authorize propagation of click event on toggle
for users that need it.


valorkin:
@H--o-l question is: if add `(click)="return false;"` should it do the same?

valorkin:
@H--o-l as far as I remember I was fighting with document click closing just opened dropdown :)

H--o-l:
@valorkin ok, not sure what you mean but I'm glad to do some tests if you have ideas.

For now I tried:
`<button dropdownToggle (click)="false">` which doesn't work and `<button dropdownToggle (click)="return false">` which doesn't compile.

Before submitting the PR I also tried to remove the `event.stopPropagation()` in dropdown-toggle directive and it was ok for my app.

Do you have other suggestions ? Or did I missed something ?

valorkin:
what if completely remove `event.stopPropagation()` from drop down toggle
but `(click)="$event.stopPropagation()"` should work in the same way
need to try this
cc #2800

H--o-l:
@valorkin just to add water under the bridge, you did say, in this discussion valor-software/ngx-bootstrap#1217, that twitter bootstrap stop the click event.
I don't know twitter bootstrap myself but it's seems a legitimate reason to put `event.stopPropagation()` where you did, but we need the option to desactivate it :-)

valorkin:
https://github.com/twbs/bootstrap/blob/v4-dev/js/src/dropdown.js#L205-L209

valorkin:
yep they use it a lot

valorkin:
@H--o-l thanks again :)
we will see where it will take us

H--o-l:
@valorkin thanks to you !

valorkin:
My hope to release it today

PR #3129, opened 2017-11-28
fix(datepicker): fixed viewed date if min max is set

### LINKED ISSUE #2711: BsDatePicker is unusable if minDate/maxDate range doesn't include current month.
If you set the minDate and maxDate to a date range that doesn't overlap with the current month (for example, a birthday for which you must be 18 or older) you can't scroll back to the valid date range because scrolling between months is disabled.  I think there are 2 issues to consider.  One that the months scrolls are disabled if you are in months that aren't included in the date range which renders the controls unusable.  However, a related issue is that when the control is opened it defaults to show the current month instead of something in the date range, and there are no configuration settings for setting what date to start by default.

### LINKED ISSUE #3123: The datepicker should display the month of [minDate] by default when opened.

When [minDate] is set dynamically (e.g. when bound to another datepicker's ngModel value) to a future date, opening the datepicker shows only greyed out dates and the user must navigate to the correct month before they can pick a date.

There is another problem here - when [minDate] is dynamically set to a date near the end of the year (e.g. December 30th 2017) it is not possible to navigate using the > button (selecting the month by name still works). This part of the bug is not reproducable in my plunkr - has it been fixed in the unpkg.com version?

[https://plnkr.co/edit/VPVQvEqu3bLwNxpsMbxx?p=preview](https://plnkr.co/edit/VPVQvEqu3bLwNxpsMbxx?p=preview)

ngx-bootstrap: 2.0.0-beta.9

Angular: 5.0.1

Bootstrap: bootstrap-sass 3.3.7

webpack

valorkin:
if no date is selected and min or max dates are excluding current month
show minimal or maximal provided date

fixes #3123 fixes #2711

### DIFF
--- a/src/datepicker/reducer/bs-datepicker.reducer.ts
+++ b/src/datepicker/reducer/bs-datepicker.reducer.ts
@@ -1,34 +1,24 @@
 // tslint:disable:max-file-line-count
-import {
-  BsDatepickerState,
-  initialDatepickerState
-} from './bs-datepicker.state';
+import { BsDatepickerState, initialDatepickerState } from './bs-datepicker.state';
 import { Action } from '../../mini-ngrx/index';
 import { BsDatepickerActions } from './bs-datepicker.actions';
 import { calcDaysCalendar } from '../engine/calc-days-calendar';
 import { formatDaysCalendar } from '../engine/format-days-calendar';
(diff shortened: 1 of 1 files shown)


YevheniiaMazur:
Tested, looks good

PR #3463, opened 2018-01-16
fix(tooltip): fix delay

### LINKED ISSUE #3038: fix(tooltip): tooltips with delay continue displaying even if not hovering after timeout
When you have multiple elements with tooltips and the tooltips have a delay, I expect the tooltips to show up after hovering the element for that specific delay time. But I expect that the tooltips don't show up when I stop hovering the element before that time is up.
But in fact the Tooltips show up if I hit them once no matter where my mouse is after that timeout.

See this example: https://stackblitz.com/edit/ngx-bootstrap-4r1ab4

Try to move the mouse really quick over many elements in the table.

ngx-bootstrap: 1.9.3

Angular: 4.3.6

But also tested it with Angular 5 and ngx.bootstrap 2.0.0-beta.8

IlyaSurmay:
Fixes #3038 

# PR Checklist

 - [ ] added/updated tests.
 - [x] added/updated demos.

### DIFF
--- a/demo/src/app/components/+tabs/demos/dynamic/dynamic.html
+++ b/demo/src/app/components/+tabs/demos/dynamic/dynamic.html
@@ -8,7 +8,6 @@
     </button>
   <hr/>
   <tabset>
-    <tab heading="Static title">Static content</tab>
     <tab *ngFor="let tabz of tabs"
          [heading]="tabz.title"
          [active]="tabz.active"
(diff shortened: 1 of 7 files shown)


demo/src/app/components/+tabs/demos/dynamic/dynamic.html
```diff
@@ -8,7 +8,6 @@
     </button>
   <hr/>
   <tabset>
-    <tab heading="Static title">Static content</tab>
```
valorkin: why did you remove it?

valorkin (changes requested):

valorkin (approved):

PR #3504, opened 2018-01-20
refactor(modal): generify BsModalRef to improve readabilty and type-checking

yankee42:
Currently this example:
```typescript
class MyComponent {
  someProperty: string;
}
modalService.show(MyComponent).content.someProperty = 'demo';
```
Compiles, but type checking for `content` ist lost. This can be fixed using a cast:
```typescript
class MyComponent {
  someProperty: string;
}
(<MyComponent>modalService.show(MyComponent).content).someProperty = 'demo';
```

This PR generifies the `BsModalRef`, so that the cast is not needed anymore and thus improves readability.

### DIFF
--- a/demo/src/app/components/+modal/demos/service-component/service-component.ts
+++ b/demo/src/app/components/+modal/demos/service-component/service-component.ts
@@ -7,7 +7,7 @@ import { BsModalRef } from 'ngx-bootstrap/modal/bs-modal-ref.service';
   templateUrl: './service-component.html'
 })
 export class DemoModalServiceFromComponent {
-  bsModalRef: BsModalRef;
+  bsModalRef: BsModalRef<ModalContentComponent>;
   constructor(private modalService: BsModalService) {}
 
   openModalWithComponent() {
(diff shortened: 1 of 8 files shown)


valorkin:
@yankee42 I suppose using default type for generics could help to avoid breaking change
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults

if it works all <null> could be removed

yankee42:
> I suppose using default type for generics could help to avoid breaking change

You are right, I added a default type and removed the generics where it was `<any>`.

I also rebased on current development. The tests fail now, but they have failed for a number of previous commits on development, so I think it is likely unrelated to my changes?

yankee42:
I rebased the PR on current `development`. The tests pass now. Let me know if there is anything missing.

daniloff200:
I see bunch of conflicts there, @yankee42 , could you please resolve them?

daniloff200:
closed in favor of #5941

Issue #3521, opened 2018-01-23
Getting "Invalid date" when using a custom dateInputFormat


FrancescoBorzi:
Might be related with https://github.com/valor-software/ngx-bootstrap/issues/3513 but I'm not sure about it.

### Bug description or feature request:

Using a custom date format, for example:

`[bsConfig]="{ dateInputFormat: 'YYYY-MM-DD' }"`

then if you try to manually passing the date value, writing the text corresponding to the date, you always get "Invalid date" even if the text represents a valid date in the format specified by `dateInputFormat`.

The bug is probably due to the fact that it still tries to parse the date according to the default date format and not according to the value specified by `dateInputFormat`.

### Plunker/StackBlitz that reproduces the issue:

https://plnkr.co/edit/8Ek11Y1d2yzyI5gPwBLb?p=preview

Steps to reproduce: try to select any date and then edit the date manually changing the text.

### Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: 2.0.1

Angular: 5.2.1

Bootstrap: 3.3.7

### Build system: Angular CLI, System.js, webpack, starter seed:

```Angular CLI: 1.6.5
Node: 8.9.1
OS: linux x64
Angular: 5.2.1
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.6.5
@angular-devkit/build-optimizer: 0.0.41
@angular-devkit/core: 0.0.28
@angular-devkit/schematics: 0.0.51
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.5
@schematics/angular: 0.1.16
typescript: 2.6.2
webpack: 3.10.0
```

bastienJS:
I am still getting invalid date having:

 dateInputFormat: 'DD.MM.YYYY'

but my assigned date is a Date object looking like: 

`2019-01-30T00:00:00+01:00`

using that configuration:

 dateInputFormat: 'DD-MM-YYYY'

the date is bound correctly.

Should I open new bug or do you guys REOPEN issues?

steffhimes:
Mine parsed when it was set to the format of the Date model. 
"2018-04-17T15:44:48" to 'MMMM Do YYYY, h:mm:ss a'  -> "invalid date".
"2018-04-17T15:44:48" to 'YYYY-MM-DD hh:mm:ss'  -> "2018-04-17 03:44:48"

PR #3522, opened 2018-01-23
Fix datepicker date parsing when using a custom date format

FrancescoBorzi:
Fixes https://github.com/valor-software/ngx-bootstrap/issues/3521

### DIFF
--- a/demo/src/app/components/+datepicker/demos/custom-format/date-picker-custom-format.html
+++ b/demo/src/app/components/+datepicker/demos/custom-format/date-picker-custom-format.html
@@ -1,0 +1,19 @@
+<div class="row">
+  <div class="col-xs-12 col-12 col-md-4 form-group">
+    <form [formGroup]="myForm">
+      <input
+        class="form-control"
+        [minDate]="minDate"
+        [maxDate]="maxDate"
+        #dp="bsDatepicker"
+        bsDatepicker
+        [formControlName]="'myDate'"
+        [bsConfig]="{ dateInputFormat: 'YYYY-MM-DD' }"
+      >
+    </form>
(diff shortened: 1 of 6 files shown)


valorkin:
Could you apply same change to date range input please

FrancescoBorzi:
done

demo/src/app/components/+datepicker/demos/custom-format/date-picker-custom-format.html
```diff
@@ -0,0 +1,19 @@
+<div class="row">
+  <div class="col-xs-12 col-12 col-md-4 form-group">
+    <form [formGroup]="myForm">
+      <input
+        class="form-control"
+        [minDate]="minDate"
+        [maxDate]="maxDate"
+        #dp="bsDatepicker"
+        bsDatepicker
+        [formControlName]="'myDate'"
```
valorkin: it's just a string setter `formControlName="myDate"`

valorkin (dismissed):

demo/src/app/components/+datepicker/demos/custom-format/date-picker-custom-format.html
FrancescoBorzi: it's just a matter of codestyle, anyway I changed it as you suggested

YevheniiaMazur:
tested, looks good

olezt:
I think there is still an issue with this. Before the fix if dateInputFormat was 'DD-MM-YYYY', entering manually date 01-02-2018 would become 02-01-2018. Now it is not reversed on input but when opening the datepicker I still see the date reversed.

PR #3573, opened 2018-01-29
docs(dropdowns): add manual showing/hiding demo

### LINKED ISSUE #3568: dropdown hide function
How to call hide function of dropdown ?
There only said dropdown has this function but not example to show it.

I use autoClose =true and preventDefault function and  stopPropagation function  to do closed outside.
This method let me can click item to call other function example customize click.

But dropdown  can not auto closed after list item click function .

I want to use hide function close it by myself to solve this problem.

artolshansky:
close #3568

# PR Checklist

 - [x] added/updated demos.

### DIFF
--- a/demo/src/app/components/+dropdown/dropdown-section.list.ts
+++ b/demo/src/app/components/+dropdown/dropdown-section.list.ts
@@ -68,6 +68,11 @@ export const demoComponentContent: ContentSection[] = [
         anchor: 'triggers-manual',
         component: require('!!raw-loader?lang=typescript!./demos/triggers-manual/triggers-manual.ts'),
         html: require('!!raw-loader?lang=markup!./demos/triggers-manual/triggers-manual.html'),
+        description: `<p>Dropdown can be triggered in two ways:
+          <ul>
+            <li>by toggling <code>isOpen</code> property</li>
+            <li>by <code>show</code>/<code>hide</code> methods from directive</li>
+          </ul></p>`,
         outlet: DemoDropdownTriggersManualComponent
       },
       {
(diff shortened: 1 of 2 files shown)


IlyaSurmay (dismissed):

IlyaSurmay (commented):
Please add a little note that informs user that he can use two types of manual triggering - via input property and directive methods

demo/src/app/components/+dropdown/dropdown-section.list.ts
```diff
@@ -68,6 +68,11 @@ export const demoComponentContent: ContentSection[] = [
         anchor: 'triggers-manual',
         component: require('!!raw-loader?lang=typescript!./demos/triggers-manual/triggers-manual.ts'),
         html: require('!!raw-loader?lang=markup!./demos/triggers-manual/triggers-manual.html'),
+        description: `<p>Dropdown can be triggered for two ways:
```
IlyaSurmay: for -> in

demo/src/app/components/+dropdown/dropdown-section.list.ts
```diff
@@ -68,6 +68,11 @@ export const demoComponentContent: ContentSection[] = [
...
+          <ul>
+            <li>by <code>toggleDropdown</code> method</li>
```
IlyaSurmay: user may not have `toggleDropdown` method because it's only defined in our demo. You should point out that user needs to manage dropdown's `isOpen` input property

IlyaSurmay (changes requested):

demo/src/app/components/+dropdown/dropdown-section.list.ts
artolshansky: changed

demo/src/app/components/+dropdown/dropdown-section.list.ts
artolshansky: changed

IlyaSurmay (approved):

YevheniiaMazur:
Tested, looks good

PR #3604, opened 2018-01-31
Fix sidebar-content display bug in the demo page

justforuse:
It's about some display bug in https://valor-software.com/ngx-bootstrap/#/getting-started
When the browser width under 1504px the sidebar-content max-height not changed correspondingly, so we can not see the Typehead menu. And the sidebar-content is a little taller when browser width above 1504px.

### DIFF
--- a/demo/src/assets/css/style.scss
+++ b/demo/src/assets/css/style.scss
@@ -539,7 +539,7 @@ a:hover {
 }
 /* below 1504 */
 @media (max-width: 1504px) {
-  .sidebar-content{
+  .sidebar-content {
     max-height: calc(100vh - 246px); // header + search field + bootstrap buttons field
   }
 }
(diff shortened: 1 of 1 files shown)


EvilAlexei:
Hello, @justforuse.
Can you provide additional info about bug?
OS, browser, example resolution and a screenshot? Screenshot will be perfect :)
Thank you!

justforuse:
@EvilAlexei 
OS: win7
Browser: Chrome64
This is a screenshot when browser width below 1504px, what I see is I can not see Typeahead menu
below1504
This is  a screenshot when browser width above 1504px, The menu not display completely.
above1504

EvilAlexei:
Thank you for your solution, but we decided to change design of buttons to prevent height changes, it will be more stable. 

Just FYI, i'm about this PR #3780

Issue #3300, opened 2017-12-15
fix(demo datepicker): when date- or datepangepicker popup is open, page under it still scrollable at mobile devices


YevheniiaMazur:
### Bug description:
- Open ngx-bootstrap demo at mobile device or mobile simulator
- Go to datepicker demo page
- Open any date- or daterangepicker popup
- Scroll page

#### Expected result:
Two cases were proposed: 
- Close datepicker popup when scroll started
- Lock scroll while datepicker popup is opened

![datepickerscrollissue](https://user-images.githubusercontent.com/27342505/34037854-f088b000-e192-11e7-8236-96f0c2569a94.gif)
### Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: all supported

Angular: all supported

Bootstrap: all supported

mridul16g:
I am still facing this issue

raideltorres:
same for me here, still facing this issue and not only in mobile or in a modal, if the datepicker is inside any scrollable container this will happen.

PR #3610, opened 2018-02-01
fix(datepicker): remove position fixed

tarusin:
related to #3300

# PR Checklist

 - [x] added/updated demos.


src/datepicker/bs-datepicker.component.ts
```diff
@@ -173,6 +173,10 @@ export class BsDatepickerDirective implements OnInit, OnDestroy, OnChanges {
         this.hide();
       })
     );
+
+    this._outsideScrollListener = this._renderer.listen('document', 'scroll', () => {
```
IlyaSurmay: Isn't this supposed to be in another PR? Also, this behavior should be disabled by default and this listener should only be created when corresponding input is set to true. input example `@Input() hideOnScroll = false`;

src/datepicker/bs-datepicker.scss
```diff
@@ -511,14 +511,6 @@ bs-daterangepicker-container {
 
 /* screen size < 1024px */
 @media (max-width: 768px) {
-  bs-datepicker-container,
-  bs-daterangepicker-container {
-    position: fixed !important;
-    left: 50% !important;
```
IlyaSurmay: I thought we were only going to remove `position: fixed` and leave top and left

IlyaSurmay (changes requested):

demo/src/app/components/+datepicker/datepicker-section.list.ts
```diff
@@ -63,6 +64,16 @@ export const demoComponentContent: ContentSection[] = [
         `,
         outlet: DemoDatePickerCustomFormatComponent
       },
+      {
+        title: 'Manual datepicker',
```
IlyaSurmay: This name doesn't reflect what demo is actually doing. Correct name is `Hide on scroll`

IlyaSurmay (changes requested):

IlyaSurmay (dismissed):

IlyaSurmay (dismissed):

IlyaSurmay (approved):

YevheniiaMazur:
Tested, looks good

PR #3678, opened 2018-02-06
keep invalid text in date input

arins:
# PR Checklist

 - [x ] read and followed the [CONTRIBUTING.md](https://github.com/valor-software/ngx-bootstrap/blob/development/CONTRIBUTING.md) guide.
 - [] added/updated tests.
 - [] added/updated API documentation.
 - [ ] added/updated demos.


src/datepicker/bs-datepicker-input.directive.ts
```diff
@@ -67,7 +68,14 @@ export class BsDatepickerInputDirective
     const initialDate = !value ? ''
       : formatDate(value, this._picker._config.dateInputFormat, this._localeService.currentLocale);
 
-    this._renderer.setProperty(this._elRef.nativeElement, 'value', initialDate);
+			if(this.keepInvalidInputValue && !isDateValid(value))
+			{
+				this._renderer.setProperty(this._elRef.nativeElement, 'value', this._elRef.nativeElement.value);
+			}
+			else
```
valorkin, line 75: may be we shouldn't do anything here in else?

and please please format if:
```
if (..) {
////
}
```

src/datepicker/bs-datepicker-input.directive.ts
```diff
@@ -39,8 +39,9 @@ export class BsDatepickerInputDirective
   private _onChange = Function.prototype;
   private _onTouched = Function.prototype;
   private _validatorChange = Function.prototype;
-  private _value: Date;
-
+	private _value: Date;
+	@Input()
+  public keepInvalidInputValue:boolean = false;
```
valorkin, line 44: `@Input() keepInvalidInputValue = false;`

valorkin (changes requested):
I suppose it could be `keepInvalidValue` ?
Plus, please, apply same changes to date range picker
Thanks in advance

valorkin:
+ I suppose a separate demo for this use case is must :)

ludmilanesvitiy:
@arins Hello! Thank you for your contribution. As I see, a few changes requested to your PR, but they are not applied. Do you still need this PR?

arins:
I am not working in that project anymore so I don't know if it is needed. I personally don't need it but maybe someone else needs it. If you don't have use for it you can close the PR

ludmilanesvitiy:
@arins OK, Thank you.
Due to absent audience activity for the proposed change, I'll close this PR. If you need something similar - don't hesitate to create a new issue/PR.

szh:
I have a need for this. Can this PR be re-opened?

szh:
Is there a workaround that works for now? I've tried adding a `bsValueChange` event listener as follows:
```
@ViewChild("dateInput") private dateInput: ElementRef;

private dateValChanged(val: Date): void {
    if (val instanceof Date && isNaN(val.getTime())) { // Invalid date
      // By default the input will show "Invalid Date", so manually set it to show the actual value
      setTimeout(() => this.renderer.setProperty(this.dateInput.nativeElement, 'value', this.field.Value), 0);
    }
  }
```
This appears to work when prefilling a field with an invalid date value, but not when manually entering it.

charlesc32:
@szh I got your solution to work with 1 additional change. Unfortuenatly you have to keep track of the value as it changes. I added a handler for keyup and stored the value. Then check that value on (instead of valueChange) and reset the input value if its invalid. It seems to work ok so far.

```
onBlur() {
        if (isNaN((new Date(this.changedValue)).getTime())) { // Invalid date
            // By default the input will show "Invalid Date", so manually set it to show the actual value
            setTimeout(() => this.renderer.setProperty(this.dateInput.nativeElement, 'value', this.changedValue), 0);
          }
    }

    onKeyup($event) {
        if ($event.currentTarget.value.toLowerCase() !== 'invalid date') {
            this.changedValue = $event.currentTarget.value;
        }
    }
```

PR #3695, opened 2018-02-06
adding id for tag

anil-kk:
adding id for tag similar to month component

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


src/datepicker/yearpicker.component.ts
```diff
@@ -32,7 +32,7 @@ import { DatePickerInnerComponent } from './datepicker-inner.component';
   </thead>
   <tbody>
     <tr *ngFor="let rowz of rows">
-      <td *ngFor="let dtz of rowz" class="text-center" role="gridcell">
+      <td *ngFor="let dtz of rowz" class="text-center" role="gridcell" id="{{dtz.uid}}">
```
valorkin: why should this be added?
plus it would be better to use `[attr.id]="dtz.uid"`

anil-kk: @valorkin I agree with you to use [attr.id]
Probably I should reflect the same changes in existing code for month picker components template.

1.) both month and date picker from ngx-bootstrap has ID attribute at this level, I guess the code in year picker already covered this case to provide unique id, why not use it? probably some one missed this case before?

2.) #3704 In our code we temporarily built outsideclick functionality. To safely determine that the click is inside content, we require unique feature, so attribute ID over an element at this level would do fine.

just let me know your thoughts

valorkin:
@anil-kk shouldnt't broke anything :)

valorkin:
Thanks for contribution

YevheniiaMazur:
Tested, looks good

valorkin (approved):

valorkin:
Thanks :)

PR #3728, opened 2018-02-08
fix(typeahead): incoming data are not filtered after typeahead kicks-in

### LINKED ISSUE #3725: fix(typeahead): incoming data are not filtered after typeahead kicks-in
# Bug description:
The list is filtered according to the entered data using user's functionality in demo, not ngx-bootstrap component's functionality.

### Steps to reproduce:
- Open async demo for typeahead
- Type one of the states, which are in the list
- See demo code

# Actual result:

The list is filtered according to the entered data using user's functionality in demo
screenshot from 2018-02-08 13-01-57

# Expected result:

The list is filtered according to the entered data using ngx-bootstrap component's functionality

# Versions of Angular, and Bootstrap:
Angular: all supported
Bootstrap: all supported

IraErshova:
Fix filtering incoming data after typeahead kicks-in. Remove filtering functionality from demo code to ngx-bootstrap component's code.

Close #3725

# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.

### DIFF
--- a/src/typeahead/typeahead.directive.ts
+++ b/src/typeahead/typeahead.directive.ts
@@ -22,7 +22,7 @@ import { TypeaheadContainerComponent } from './typeahead-container.component';
 import { TypeaheadMatch } from './typeahead-match.class';
 import { TypeaheadConfig } from './typeahead.config';
 import { getValueFromObject, latinize, tokenize } from './typeahead-utils';
-import { debounceTime, filter, mergeMap, switchMap, toArray } from 'rxjs/operators';
+import { debounceTime, filter, map, mergeMap, switchMap, toArray } from 'rxjs/operators';
 
 @Directive({selector: '[typeahead]', exportAs: 'bs-typeahead'})
 export class TypeaheadDirective implements OnInit, OnDestroy {
@@ -53,7 +53,7 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
    * contains the group value, matches are grouped by this field when set.
    */
   @Input() typeaheadGroupField: string;
-  /** should be used only in case of typeahead attribute is array.
+  /** should be used only in case of typeahead attribute is Observable of array.
    * If true - loading of options will be async, otherwise - sync.
    * true make sense if options array is large.
    */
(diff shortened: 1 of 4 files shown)


src/typeahead/typeahead.directive.ts
```diff
@@ -344,7 +344,15 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
     this._subscriptions.push(
       this.keyUpEventEmitter
         .debounceTime(this.typeaheadWaitMs)
-        .mergeMap(() => this.typeahead)
+        .mergeMap((value: string) => {
+          const normalizedQuery = this.normalizeQuery(value);
+
+          return this.typeahead.map((data: any[]) => {
```
valorkin: `any`? - use types or interfaces
remove `return`s

src/typeahead/typeahead.directive.ts
```diff
@@ -344,7 +344,15 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
     this._subscriptions.push(
       this.keyUpEventEmitter
         .debounceTime(this.typeaheadWaitMs)
-        .mergeMap(() => this.typeahead)
+        .mergeMap((value: string) => {
```
valorkin: `typeahead` can be almost anything, and you would knew about it, if we would have type on it :(
so this PR will break 99% of use cases

valorkin (changes requested):

src/typeahead/typeahead.directive.ts
IraErshova: `value: string` is not typing for `typeahead`. This code is correct and doesn't break anything

src/typeahead/typeahead.directive.ts
IraErshova: `typeahead` has type any in the whole module. Adding type for this instance should be a separate issue.

daniloff200 (dismissed):

dmitry-zhemchugov:
Tested. Issues were fixed by @IraErshova . Ready to merge

daniloff200 (approved):

valorkin (approved):

valorkin:
@dmitry-zhemchugov has to be tested with real XHR as a source of data

Issue #3544, opened 2018-01-25
Timepicker: Placeholders are not editable


TomONeill:
The Timepicker has placeholders in the inputs (HH, MM, SS) that are hardcoded in the template. HH, MM and SS are not valid in the language I'm developing for, so I need to be able to turn those off.

Could you look into the possibility of changing the placeholders manually or perhaps turn them off entirely?

Thank you in advance.

https://github.com/valor-software/ngx-bootstrap/blob/development/src/timepicker/timepicker.component.html

urstrulykkr:
Hi @daniloff200 
I would like to work on this issue.

daniloff200:
@urstrulykkr  looks like, we have a PR from @fhuschle  already.

urstrulykkr:
@daniloff200 , Can you please assign me to any beginner level hacktoberfest issue?
Thanks in advance :)

daniloff200:
@urstrulykkr you can try to pick https://github.com/valor-software/ngx-bootstrap/issues/5401 , I added a description, what has to be done, so, I think it's good to go! 
Feel free to ask any questions in this issue, if you have any

PR #3787, opened 2018-02-15
Timepicker: Added showPlaceholders boolean. Ref #3544

TomONeill:
Added an extra input for the Timepicker "showPlaceholders" to show/hide the placeholders like requested in Issue #3544


valorkin:
@TomONeill hey, I am agreed with idea that hardcoded placeholder is a bad idea
but adding such check in template affects perf

maybe you could add configuration of placeholders text to TimepickerConfig?

daniloff200:
I think, that can be closed in favor of #5429

PR #3894, opened 2018-02-28
Bugfix/tab select

arpadvas:
# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.

There is an interference between the tab select event emitter and the select browser event. So if you have an input or textarea html element inside the tab and you select the value, it fires the function listening to tab select event. So I've changed the select and deselect output names to selectTab and deselectTab which prevents the issue form happening.


valorkin:
hey, thanks for PR!
That sucks, you will have to stop propagation for now,
or need to take a look into event target
cuz such change is a breaking change and major version bump will be required to publish it

ludmilanesvitiy:
This fix is already merged via other PR: https://github.com/valor-software/ngx-bootstrap/pull/5002

PR #3907, opened 2018-03-01
fix(timepicker): remove unnecessary space when showMeridian is false

s00pa:


valorkin (approved):

valorkin:
Thanks for idea! 
And congratulations with first contribution :)

YevheniiaMazur:
Tested, looks good

Issue #3623, opened 2018-02-01
Add [typeahead] directive selector prefix


anjmao:
Hi,
I'm developing dropdown select called `ng-select` and there is one issue https://github.com/ng-select/ng-select/issues/224
It looks like there is `typeahead` [directive](https://github.com/valor-software/ngx-bootstrap/blob/development/src/typeahead/typeahead.directive.ts#L32) which use selector as `[typeahead]` and this collides with `ng-select` input called `typeahead`.

Is it possible to solve this problem without changing prefix of valor-software typeahead directive? In any way it is a good practice to add prefix to such directives.

andjiev:
@IlyaSurmay Seems like this isn't fixed. I have the same problem.

kunovskis:
I have the same problem. Error limited.map is not a function in console.

valorkin:
@anjmao it will have a prefixed version in next major version (with angular 6)
and removed in next+1 major version (with angular 7)
so it will be easier to prefix input it `ng-select`

markleusink:
Have the same issue. Did the PR already make it into a release?

bogdan-melov:
I have a console error cuz of this. Please fix it

williamverdolini:
Why this issue is closed?
The PR is not in master, but even if it will be merged, the PR will no fix the issue, cause "the [typeahead] selector still exists" as the author has written

Please reopen it

g3344:
Did anyone find an answer to this?  Running into the same issue

mtgleeson:
I'm hitting the same issue - any fix? 
I'm using ngx-bootstrap 6.2.0 and @ng-select/ng-select 6.1.0.

PR #4033, opened 2018-03-20
Added prefix for typeahead directive

andjiev:
#3623 

Since more people have encountered the same problem with the usage of the `typeahead` directive, I've made a pull request which solves that problem by adding a prefix to the directive. Because changing the selector will cause a breaking change, I added the selector with the prefix while keeping the old one for now and marking it as deprecated. This is not a fix for us because the `[typeahead]` selector still exists, but we hope that the selector will be removed and switched with our new one in the next release.


valorkin:
Thanks for PR!
closing cuz of: https://github.com/valor-software/ngx-bootstrap/issues/3623#issuecomment-377193968

PR #4096, opened 2018-03-26
docs(tooltip): add aria attributes to tooltip demo component

### LINKED ISSUE #4094: docs(tooltip): add aria-attributes to tooltip demo component
Add aria-attributes for tooltip: 
- aria-describedby to button, which opens tooltip
(The role has already been added)

source: https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

ghost:
Was added aria-attribute to tooltip demo component:
 - `aria-describedby` to a button, which shows a tooltip
The tooltip has an id. The value of the id and value of the `aria-describedby` (button) need to be the same.

joxi_screenshot_1522673294502

Closes #4094

### DIFF
--- a/src/tooltip/tooltip-container.component.ts
+++ b/src/tooltip/tooltip-container.component.ts
@@ -43,6 +43,7 @@ export class TooltipContainerComponent implements AfterViewInit {
   placement: string;
   containerClass: string;
   animation: boolean;
+  id: string;
 
   get isBs3(): boolean {
     return isBs3();
--- a/src/tooltip/tooltip-container.component.ts
+++ b/src/tooltip/tooltip-container.component.ts
@@ -14,6 +14,7 @@ import { isBs3 } from 'ngx-bootstrap/utils';
     '[class]':
       '"tooltip in tooltip-" + placement + " " + "bs-tooltip-" + placement + " " + placement + " " + containerClass',
     '[class.show]': '!isBs3',
+    '[attr.id]': 'this.id',
     role: 'tooltip'
   },
   styles: [
--- a/src/tooltip/tooltip.directive.ts
+++ b/src/tooltip/tooltip.directive.ts
@@ -2,7 +2,8 @@
 import {
   Directive,
   ElementRef,
-  EventEmitter, HostBinding,
+  EventEmitter,
+  HostBinding,
   Input,
   OnDestroy,
   OnInit,
@@ -189,6 +190,8 @@ export class TooltipDirective implements OnInit, OnDestroy {
     this.triggers = (value || '').toString();
   }
 
+  @HostBinding('attr.aria-describedby') ariaDescribedby = `tooltip-${this.tooltipId}`;
+
   /** @deprecated */
   @Output()
   tooltipStateChanged: EventEmitter<boolean> = new EventEmitter<boolean>();
@@ -197,8 +200,6 @@ export class TooltipDirective implements OnInit, OnDestroy {
   protected _tooltipCancelShowFn: Function;
 
   private _tooltip: ComponentLoader<TooltipContainerComponent>;
-
-  @HostBinding('attr.aria-describedby') ariaDescribedby = `tooltip-${this.tooltipId}`;
   constructor(
     _viewContainerRef: ViewContainerRef,
     private _renderer: Renderer2,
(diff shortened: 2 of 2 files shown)


IlyaSurmay (changes requested):
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute

MDN says that `aria-describedby` attribute must contain an ID of the element with description. In this PR you just have text like `showTooltipOnHover` and so on. This isn't an ID of the element so there's no sense in it.

W3 ARIA docs says the same:
The element that serves as the tooltip container has role tooltip.
The element that triggers the tooltip references the tooltip element with aria-describedby.

src/tooltip/tooltip.directive.ts
```diff
@@ -12,11 +12,14 @@ import { parseTriggers } from '../utils/triggers';
...
 import 'rxjs/add/observable/timer';
 
+let newId = 0;
+
 @Directive({
   selector: '[tooltip], [tooltipHtml]',
   exportAs: 'bs-tooltip'
 })
 export class TooltipDirective implements OnInit, OnDestroy {
+  private idTooltip = newId++;
```
IlyaSurmay: more likely `tooltipId` or `id`

src/tooltip/tooltip-container.component.ts
```diff
@@ -42,6 +42,8 @@ export class TooltipContainerComponent implements AfterViewInit {
   containerClass: string;
   animation: boolean;
 
+  @HostBinding('attr.id') ariaDescribedby: string;
```
IlyaSurmay: attribute `id` is bound to `ariaDescribedby`? doesn't seem very logic to me. change property name to `id` or `containerId`

IlyaSurmay (changes requested):

src/tooltip/tooltip.directive.ts
```diff
@@ -248,7 +253,8 @@ export class TooltipDirective implements OnInit, OnDestroy {
         .show({
           content: this.tooltip,
           placement: this.placement,
-          containerClass: this.containerClass
+          containerClass: this.containerClass,
+          ariaDescribedby: this.ariaDescribedby
```
IlyaSurmay: id: this.ariaDescribedby

src/tooltip/tooltip-container.component.ts
```diff
@@ -14,6 +14,7 @@ import { isBs3 } from '../utils/theme-provider';
     '[class]':
       '"tooltip in tooltip-" + placement + " " + "bs-tooltip-" + placement + " " + placement + " " + containerClass',
     '[class.show]': '!isBs3',
+    '[attr.id]': 'this.ariaDescribedby',
```
IlyaSurmay: '[attr.id]': 'this.id',

IlyaSurmay (changes requested):

src/tooltip/tooltip.directive.ts
ghost: fixed

src/tooltip/tooltip.directive.ts
ghost: fixed

src/tooltip/tooltip-container.component.ts
ghost: fixed

src/tooltip/tooltip-container.component.ts
ghost: fixed

src/tooltip/tooltip.directive.ts
```diff
@@ -2,7 +2,7 @@
 import {
   Directive,
   ElementRef,
-  EventEmitter,
+  EventEmitter, HostBinding,
```
IlyaSurmay: please keep import syntax consistent, move `HostBinding` to a new line

IlyaSurmay (changes requested):

src/tooltip/tooltip.directive.ts
ghost: fixed

Domainv (dismissed):

IlyaSurmay (dismissed):

ludmilanesvitiy:
Tooltip demo tested in CHrome ,FF - visually nothing changes, html contain appropriate attributes and ID's. RFM

valorkin (approved):

PR #4175, opened 2018-04-03
Fix console error spam while running unit tests

FrancescoBorzi:
Unit test output with this change:

image

Current unit test output:

image


valorkin:
okay :)
thanks for PR

PR #4242, opened 2018-04-17
Allow modals for templates with context

julianseeger:
Currently, when calling modalService.show with a TemplateRef, the initialState is ignored.
With this change, the template gets its context and code like the following is possible:
```HTML
... (click)="openModal(template, {someproperty: 'with value'})"...

<ng-template #template let-someproperty="someproperty">
    {{someproperty}}
</ng-template>
```

**edit:** is there any interest in this change?

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


valorkin:
Looks good to me :)

daniloff200 (approved):

daniloff200:
I've tried case with this code, and pass data to the modal, got an error 

Here's the modal code: 

 

```
<button type="button" class="btn btn-primary" (click)="openModal(template, {somedata: 'value'})">Create template modal</button>
  <ng-template #template let-somedata="somedata">
    <span> {{somedata}} </span>
 </ng-template>
```

 

Selection_273

elliotwesoff:
I get the same error as @daniloff200 . This feature would be extremely helpful if it worked! I don't want to create components for every template that needs to take advantage of input parameters. Is there any chance this could be fixed?

julianseeger:
Nope sorry. IIRC it worked for me perfectly 3 years ago but when I checked the code today, I even had problems following the control flow to that line.  
It looks like the context doesn't even reach that point now.
The Project where this issue came up is moving to a simpler approach for modals meanwhile and it doesn't look like I'll spend enough time on this problem to fix it. So I'm really sorry but if nobody else fixes this, feel free to close this PR.  
If it wasn't important enough for the last 3 years it probably isn't now.

elliotwesoff:
@julianseeger the HEAD on that file shows that the context and initialState variables should be available in the scope of that function, but which is the right one to use? I don't mind fixing it myself and creating a PR, but following the control flow is quite difficult and I can't figure out how to debug it locally. @daniloff200 help!

daniloff200:
Well, I've tried to use these changes again, but still, got the same error.
Not sure, is it a correct way to fix that, or not

elliotwesoff:
@daniloff200 well that's a bummer. Thanks for your help anyway!

PR #4268, opened 2018-04-26
feat(typeahead): enable/disable options to make selection on tab-event (#3842)

### LINKED ISSUE #3842: feat:typeahead Make selection on TAB optional
Hi Folks,

we are using ngx-bootstrap 1.9.3 and wanted to update to 2.0.2, but what keeps us from doing so, is the fact that Typeahead selects the highlighted Option on TAB-Button.

Our Feature-Request:
Would it be possible to change the "select on tab" that was handled in [Issue 3106](https://github.com/valor-software/ngx-bootstrap/issues/3106) to be optional?

Maybe add a new Input "disableSelectOnTab" of type boolean to the component and if its not set/false, the TAB Key triggers a selection (as it does now). If you would set "disableSelectOnTab" true then the Typeahead would behave like it did before and only select on "Enter"-Key.

What do you guys think of this?

SidekickJohn:
Adds an optional Input to typeahead directive. Initially true. If set to false, tab will just do its default behaviour.

 Closes #3842

# PR Checklist

 - [X] read and followed the [CONTRIBUTING.md](https://github.com/valor-software/ngx-bootstrap/blob/development/CONTRIBUTING.md) guide.
 - [ ] added/updated tests.
 - [X] added/updated API documentation.
 - [X] added/updated demos.


src/typeahead/typeahead.directive.ts
```diff
@@ -73,8 +73,11 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
   /** should be used only in case typeaheadSingleWords attribute is true.
    * Sets the word delimiter to match exact phrase.
    * Defaults to simple and double quotes.
-   */
+   */  
   @Input() typeaheadPhraseDelimiters = '\'"';
+  /** used to enable/disable the selection on TAB Event.   
+   */
+  @Input() typeaheadSelectOnTab: boolean = true;
```
valorkin: Hey, thanks for PR, all looks good.
But one change is must for merge,
in terms of app consistency and perf, it would be better to have `selectOnTab` in a config
and not in `@Input`

valorkin (changes requested):

src/typeahead/typeahead.directive.ts
SidekickJohn: Yes you are right. But same goes for "typeaheadSingleWords" and other boolean values... 
Still: I'm trying to change the `selectOnTab` to be a config and will update my PR.
@valorkin is there an example (for the config part) that I could learn from? Or does the Typeahead-directive already have a config somewhere?

SidekickJohn: @valorkin @IlyaSurmay I still need a little advice on how to solve this one as a config. Is there an example on how to put things like this in a config?

valorkin:
As a heads up, I remember about this PR

SidekickJohn:
Ok, I made the changes. Hope I did the right things. @valorkin could you please check the PR again?

src/typeahead/typeahead.directive.ts
```diff
@@ -77,7 +78,13 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
   @Input() typeaheadPhraseDelimiters = '\'"';
   /** used to enable/disable the selection on TAB Event.   
    */
-  @Input() typeaheadSelectOnTab: boolean = true;
+  @Input()
```
valorkin: Don't rename the field, just set it with assigment
Field rename will break ppl's code

src/typeahead/typeahead.directive.ts
```diff
@@ -73,8 +74,17 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
   /** should be used only in case typeaheadSingleWords attribute is true.
    * Sets the word delimiter to match exact phrase.
    * Defaults to simple and double quotes.
-   */
+   */  
   @Input() typeaheadPhraseDelimiters = '\'"';
+  /** used to enable/disable the selection on TAB Event.   
+   */
+  @Input() 
+  get selectOnTab(): boolean {
```
valorkin, line 82: I believe this option is fine to be only in config, no need for additional input

src/typeahead/typeahead.directive.ts
```diff
@@ -140,12 +151,14 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
...
               cis: ComponentLoaderFactory,
-              private changeDetection: ChangeDetectorRef) {
+              private changeDetection: ChangeDetectorRef,
+              config: TypeaheadConfig) {
     this._typeahead = cis.createLoader<TypeaheadContainerComponent>(
       element,
       viewContainerRef,
       renderer
     );
+    Object.assign(this, config);
```
valorkin, line 161: set it to `this._config`

valorkin (changes requested):
This great, almost there!
Couple small changes left.

Each @Input adds performance cost, I believe there are no need for input for this option

src/typeahead/typeahead.directive.ts
SidekickJohn: @valorkin: How can a User of the typeahead-directive chose whether the option is selectOnTab = true or false? When putting the directive on an input-element, how can i decide this when there is no ` @Input()` for it? <br/>
Should I do it like the Datepicker Component and implement a setConfig() Method that reassigns the _config? Then there would still need to be an `@Input()` for the config...

SidekickJohn: @valorkin @Domainv I still don't know what to do right here. For us it is required that the user can decide whether the selectOnTab is true or false in the HTML for each typeahead instance seperately. Could you please be a little more precise what you expect here? Or tell me an example file in an other component where this is solved the way you imagine it.

ludmilanesvitiy:
This feature is already implemented and presented on the demo page:
https://valor-software.com/ngx-bootstrap/#/typeahead#selected-first-item

PR #4366, opened 2018-05-21
treat rxjs as externals in es2015 bundle

### LINKED ISSUE #4364: Class constructor cannot be called without 'new' when using es2015
### Bug description:
I tried to use es2015 compilation target and got an error "Class constructor cannot be called without 'new'" here:

```
class MiniState extends BehaviorSubject {
    constructor(_initialState, actionsDispatcher$, reducer) {
        super(_initialState);
^^^^^^^^^^^^^^^^^ Fails here
        const actionInQueue$ = actionsDispatcher$.pipe(observeOn(queue));
        const state$ = actionInQueue$.pipe(scan((state, action) => {
            if (!action) {
                return state;
            }
            return reducer(state, action);
        }, _initialState));
        state$.subscribe((value) => this.next(value));
    }
}

```

I believe the issue is the fact that BehaviorSubject is written in ES5 style. 
You package ES5 RxJS together with ES6 compiled ngx-bootstrap.
Though you shouldn't package RxJS at all, it should be completely external.

ngx-bootstrap:3.0.0
Angular: 6.0.2
Bootstrap:3.3.7

Webpack 4.8.3

sherlock1982:
closes #4364


valorkin:
@sherlock1982 hey, how it goes? did this work for you?

sherlock1982:
Well it made bundle smaller and now rxjs externals are properly treated as externals. Before some of them were inbuilt.

But due to overall issues with ES2015 (issue with DevBuild optimizer, issue with UglifyJS ) I can't properly test it. 
In the end I'd like to have it as separate modules for tree shaking, not like one bundle. But it's yet to come.

valorkin:
@sherlock1982 then I am closing this PR
PS right now I am adding bazel build here :)
should solve this issues, but for sure will add some new :D

Issue #2356, opened 2017-08-04
fix(modal): modal does not properly work in a lazy loaded module


dzonatan:
To have working modal in a lazy-loaded module you have to import modal module with `forRoot()` in that lazy module which is not correct as `BsModalService` will get a new instance and will not know anything about other modals of other modules (e.g. nesting won't properly work). If you skip the `forRoot()` you will get:

```
ERROR Error: No component factory found for MyModalInLazyLoadedModule. 
Did you add it to @NgModule.entryComponents?
```

That's happening because [ComponentLoaderFactory](https://github.com/valor-software/ngx-bootstrap/blob/development/src/modal/modal.module.ts#L17) is provided only in root module which does not know anything about lazy-loaded injector and factories.

Proper fix would be to have some kind of [overlay service](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal.ts#L47)  that is provided with every [ModalModule import](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal.module.ts#L15).

Reproduction - https://stackblitz.com/edit/ngx-bootstrap-playground-mtwpzg

nischi:
Is there a workaround? I have exactly the same issue

**Edit**
I have a workaround. I will use the directive with some ViewChilds.

psined1:
Running into the same issue. Could you share an example please?

dzonatan:
@psined1 Just import `ModalModule.forRoot()` in your lazy-loaded module.

psined1:
yep, did that, works fine

julianonunes:
Also having the same issue. The workaround pointed by @dzonatan fixes it, but I don't know what are the implications of using forRoot in a child module.

Strike117:
Having same issue, is there an update?

vincentjames501:
I wouldn't reimport `ModalModule.forRoot()` as it will create a new BsModalService and then you'll have multiple of these services which are trying to manage some global UI state. It would work fine if you don't plan on using the service to call something like `bsModalService.hide(1)` from another context (which our app does to force close modals upon a navigation event).

[...]

PR #4370, opened 2018-05-21
Allow showing modals using components from lazy loaded module

vincentjames501:
A potential way to address #2356. The issue with using `.show()` and trying to show a component from a lazy loaded module is that the `ComponentLoaderFactory` created from `ModalModule.forRoot()` can only resolve components from the root injector and not from within the child injector. If the lazy loaded module provides its own `ComponentLoaderFactory`, it will receive a `ComponentFactoryResolver`/`Injector` that can properly resolve the lazy loaded module's entryComponents, providers, etc.

Another options would be to supply the `ComponentFactoryResolver` directly instead of the `ComponentLoaderFactory` but I think ultimately there is more flexibility in this approach.

### DIFF
--- a/demo/src/app/components/+modal/modal-section.list.ts
+++ b/demo/src/app/components/+modal/modal-section.list.ts
@@ -66,7 +66,9 @@ export const demoComponentContent: ContentSection[] = [
           in <code>.show()</code> method as in example, and don't forget to include your component to
           <code>entryComponents</code> of your <code>NgModule</code><br> If you passed a component
           to <code>.show()</code> you can get access to opened modal by injecting <code>BsModalRef</code>. Also you can pass data 
-          in your modal by adding <code>initialState</code> field in config. See example for more info</p>`,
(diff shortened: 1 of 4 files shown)


vincentjames501:
FWIW #3361 seems like a better approach than this. I hope that can get looked at soon.

valorkin:
Yep, I have faced this while upgrading to v6 of angular, and any solution will be a breaking change. I think adding Component Loader v2 should be a good choice. But I will take a look on PRs for sure

xino1010:
@vincentjames501  when will this functionality be available in production?

valorkin:
Even now you should be able to use forRoot in child module and it will work
This PR will be merged soon

valorkin:
Thanks for PR
But it doesn't fix an issue, just adds workaround
IMHO you can add `ModalModule.forRoot()` in `imports` of lazy loaded module and it will work as a hack for now

Hint: Good fix would be converting BsModalService into `useFactory`, but it will require on more service - to handle nested modal loading.

Thanks again

PR #4509, opened 2018-07-18
fix(datepicker): Fix navigation alignment with preserveWhitespaces: false

### LINKED ISSUE #4443: Datepicker header alignment
The part in bs-datepicker-navigation-view is not aligned properly. The arrows and date are not centered. 

ngx-bootstrap: 3.0.1

Angular: 6.0.5

Bootstrap: 3.3.7

adrienverge:
When used with `preserveWhitespaces: false` compilation option in
Angular, the datepicker title elements are misaligned, because spaces
are missing in the justified text.

Since Angular 6, `preserveWhitespaces: false` is the default, so many
users will encounter this bug.

Let's add zero-width spaces in order to have a correct alignment in all
cases.

Closes #4443

### DIFF
--- a/demo/src/main.ts
+++ b/demo/src/main.ts
@@ -9,4 +9,4 @@ if (environment.production) {
   enableProdMode();
 }
 
-platformBrowserDynamic().bootstrapModule(AppModule, {preserveWhitespaces: true});
+platformBrowserDynamic().bootstrapModule(AppModule);
--- a/demo/src/tsconfig.json
+++ b/demo/src/tsconfig.json
@@ -24,8 +24,5 @@
   "exclude": [
     "src/test.ts",
     "**/*.spec.ts"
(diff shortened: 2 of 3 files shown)


valorkin (dismissed):

valorkin:
@adrienverge thanks for PR! please remote white space:true for src/demo folder and push
so QA could test this change locally

adrienverge:
@valorkin Sure.

I added a commit:

---

### fix(main): Reset `preserveWhitespaces` option to default

This commit reverts fee535a and 0478637, which should not be needed
anymore.

`preserveWhitespaces: false` is the default since Angular 6: let's use
the default!

YevheniiaMazur:
Tested, looks good

valorkin (approved):

PR #4605, opened 2018-09-11
perf(dropdown): reduce the number of document click listeners

### LINKED ISSUE #2172: fix(dropdown): performance problems with multiple dropdowns

Dragsaw:
Edit document click listeners to be added only for open dropdowns which improves overall page performance in case of a large number of dropdowns.

Currently all dropdowns are listening for document clicks and if there're lots of them any click results in a noticeable page-freeze. This is fixed by listening for document clicks only when a dropdown is open.

This is basically a clone of #2419, which fixes #2172. Looks like the component was rewritten, but the changes from #2419 were omitted for some reason.


charliemc:
This would be a very interesting fix to have <3

ludmilanesvitiy:
Tested with 1000 and 5000 dropdowns, looks good. Performance now is much better.

Domainv (approved):

Issue #4417, opened 2018-06-07
Typeahead breaks if typeahead property receives NULL value


ghost:
The typeahead directive breaks if the [typeahead] property receives a NULL value (with NullPointerException). This happens since my async data service returns NULL instead of [] if no search results were found.

I'm wondering why NULL is not a valid value? In my opinion, the directive should catch undefined values by itself.

```
ERROR TypeError: Cannot read property 'slice' of null
    at TypeaheadDirective.push../node_modules/ngx-bootstrap/typeahead/typeahead.directive.js.TypeaheadDirective.prepareMatches (typeahead.directive.js:306)
...
```

ngx-bootstrap: 3.0.0
Angular: 6.0.1

pankmuze:
I am also facing the same issue. When there is nothing to return from the database for the API call, it is throwing the slice error and then I am unable to search anything anymore. 

Can someone help in this?

pankmuze:
Can we use the typeaheadNoResults in anyway here?

pankmuze:
Found a workaround. For the empty return value, I am making the API return an empty object, so the slice error is not coming up anymore. 

This empty object I have renamed as "No record found", so that when it is seen in the UI, it looks as though it is not returning anything.

In the TypeaheadOnSelect method, I am clearing the input field :-)

Hope this help others who are facing this issue

PR #4611, opened 2018-09-14
fix(typeahead): fix typeahead's breaking if typeahead property receives NULL

ghost:
Related #4417


Domainv:
code coverage test is failed

Domainv:
Should be fixed inside typeahead

Issue #1884, opened 2017-04-07
fix(typeahead): typeahead focus obtained again after leaving the control


diegofal:
I am experiencing the following issue:

- I setup the control in the following way. Async call is done to an API, which hits an external resource. This operation takes, at least, 1 second. 

  <input class="form-control"
             name="searchForm"
             required
             #searchForm="ngModel"
             [(ngModel)]="searchModel"
             [typeahead]="searchResults"
             (typeaheadOnSelect)="selectFromSearchResult($event)"
             typeaheadOptionField="fullName"
             typeaheadWaitMs="200"
             (keypress)="loading=true"
             size="30" type="text" />

Component code is as follows:

 this.searchResults = Observable.create((observer:any) => {
      this.loading = true;
      this.contactsService.search(this.searchModel)
        .subscribe((result: any) => {
          this.loading = false;
          observer.next(JSON.parse(result._body));
        })
    });

- Upon typing, after 200ms a request is executed.
- I hit tab before the response comes back, so focus is now in another control.
- After the results come back from the server, typeahead control obtains focus again. This is not an expected behaviour.

Q: is there a way to cancel the request? 
Q: shouldn't the control cancel the request on focus lost? 

Thank you in advance!

jvanrossum:
+1
We are also experiencing issues caused by this behaviour

Hadev-JHH:
We have run into this issue as well. It results in some very funny behaviour when you have 2 typeaheads next to eachother where they keep stealing focus from eachother

Looking at the code it seems to come from:

1. The fact that the async actions are finalized regardless of focus state
2. The fact that finalizeAsyncActions calls show() regardless of focus state
3. The fact that show() calls focus() on the input element

lalocespedes:
Here is a issue example in stackblitz. Typeahead dont stop when lost focus.

https://stackblitz.com/edit/angular-ngx-typeahead

Help please,

PR #4622, opened 2018-09-18
fix(typeahead): fix focus after leaving control

tarusin:
Related to: #1884
Added new demo with input which demonstrates behavior with closed dropdown when we leaving input field.


Domainv:
code coverage

milosdanilov:
@tarusin @Domainv Is this going to be merged soon?

Domainv:
@milosdanilov Not sure, the PR require improvements

ejasarevic:
Any progress/plan for this PR?

daniloff200:
@EmanJasarevic oh, that's pretty old one
I hope to have time to look at it, and rework soon. Let's hope that it will be available in next or next two releases

ejasarevic:
Hi @daniloff200,
thanks for making these changes. However, IMO it still behaves differently than expected (even demo example https://valor-software.com/ngx-bootstrap/#/typeahead#cancel-on-focus-lost).
Here are a couple of use cases which I've tested, correct me if I have wrong expectations:
1. Start typing something (e.g "Wash") and wait for the async data to be loaded
Expected: Data loaded properly, all good.
Result: ✔️ 
2. Start typing "Wash" and pres tab a couple of times to move the focus from typeahead:
Expected: Typeahead will cancel async request and not load suggestions ("Washington" in this case).
Result: ➖ 
3. Start typing "Wash" and click outside of the input:
Expected: Typeahead will cancel async request and not load suggestions ("Washington" in this case).
Result: ➖ 

In our app we use the typeahead to provide suggestions, but also allow user to enter freetext in the input as well and accept it even though it is not option from the typeahead. For example, You start typing "Wash." and press enter/tab, focus will move to the next control and your input will be accepted. Is it doable with the current Typeahead implementation?

Regards

chanageller:
@daniloff200
This still seems to be broken. cancelRequestOnFocusLost is always set to false.
There seems to be a bug in the code, when assigning the variables with the config settings the naming is wrong.

line 189 in typeahead directive-

      {
        typeaheadHideResultsOnBlur: config.hideResultsOnBlur,
        typeaheadCancelRequestOnFocusLost: config.cancelRequestOnFocusLost,
        typeaheadSelectFirstItem: config.selectFirstItem,
        typeaheadIsFirstItemActive: config.isFirstItemActive,
        typeaheadMinLength: config.minLength,
        adaptivePosition: config.adaptivePosition,
        isAnimated: config.isAnimated
      }
    );`
**typeaheadCancelRequestOnFocusLost** should be **cancelRequestOnFocusLost**

daniloff200:
woah
@chanageller  can you submit a PR with that, and then we can check that?

PR #4631, opened 2018-09-21
feat(typeahead): add Input Property for selected First item in option list

### LINKED ISSUE #3965: feat:typeahead Don't select any fields until the first up/down arrow keypress
We've been having problems with the typeahead behavior introduced in https://github.com/valor-software/ngx-bootstrap/pull/2839 in conjunction with setting `minLength` to 0.

Due to the above feature, when a typeahead has a `minLength` of 0, when keyboard / screen reader users Tab from field to field, as they go through the typeahead field they unintentionally select the first item from the list.

Disabling the select on Tab feature as outlined in #3842 would be one way to fix the issue, but we _do_ like the idea behind the feature, and we think that a fairly minimal change in behavior could allow keeping the feature around while also improving the experience for keyboard users.

Our proposal is to change the initial state when the typeahead dropdown menu is first shown, so that rather than the first item being active and selected on Enter/Tab, no items are active until the first up/down arrow keypress, and so enter/tab would not select anything until then.

Would this be a welcome change? Would this be a desired default behavior or an optional one?

We're very interested in this feature (or something similar) and/or #3842, and would be happy to implement the feature ourselves, especially if it would be welcomed by the project maintainers.

ghost:
Related #3983
Closes #3965
Added to this pr an opportunity to choose: 
- the first item in the option list is selected or not.
We can do it adding a input property typeaheadSelectedFirstItem on input html element


src/typeahead/typeahead-container.component.ts
```diff
@@ -54,8 +55,11 @@ export class TypeaheadContainerComponent {
   @ViewChildren('liElements')
   private liElements: QueryList<ElementRef>;
 
-  constructor(element: ElementRef, private renderer: Renderer2) {
+  constructor(element: ElementRef,
+              private renderer: Renderer2,
+              config: TypeaheadConfig) {
```
valorkin: remove import of config

valorkin (changes requested):
revert changes to package-lock.json

first element should be highlighted but not selected on focus lost

if min length === 0 && input is empty && first element is active -> do nothing on focus lost

src/typeahead/typeahead-container.component.ts
ghost: done

Domainv (dismissed):

ludmilanesvitiy:
@Domainv THis PR tested, looks good. But PR #3983 should be reworked (for saving typeaheadMinLength demo)

valorkin (approved):

PR #4649, opened 2018-09-27
fix(timepicker): fix tslint errors

tarusin:
# PR Checklist
Before creating new PR, please take a look at a checklist below to make sure that you've done everything that needs to be done before we can merge it.


src/timepicker/reducer/timepicker.reducer.ts
```diff
@@ -36,6 +36,7 @@ export const initialState: TimepickerState = {
   }
 };
 
+// tslint:disable-next-line
```
Domainv: Which rule do we turn off?

Issue #4685, opened 2018-10-15
Typeahead: Possibility to open link in a new tab with ctrl+click


cakeinpanic:
Using typeahead I want to have possibility to open links in a new tab.
By now it it not possible at all, cause all clicks are set to preventDefault and ctrl+click or mousewheel can not be differed – the only case is to use `<a>` tag and a context menu

svetoldo4444ka:
Sorry, but a possibility to open link in a new tab is contrary to standard typeahead's behavior

valorkin:
interesting, thanks

IraErshova:
@cakeinpanic Hi! I am not sure that we are going to add this functionality to typeahead component. But you can use `typeaheadOnSelect` event to add functionality for opening any URL in a new tab. Could you please check my example in Stackblitz?
https://stackblitz.com/edit/angular-8ioart

PR #4687, opened 2018-10-15
feat(typeahead): Provide event which caused the selection

cakeinpanic:
I propose the solution for https://github.com/valor-software/ngx-bootstrap/issues/4685 – I suggest providing event(MouseEvent or KeyboardEvent) with emitted TypeaheadMatch

### DIFF
--- a/src/typeahead/typeahead-container.component.ts
+++ b/src/typeahead/typeahead-container.component.ts
@@ -100,8 +100,8 @@ export class TypeaheadContainerComponent {
     return this.parent ? this.parent.typeaheadItemTemplate : undefined;
   }
 
-  selectActiveMatch(): void {
-    this.selectMatch(this._active);
+  selectActiveMatch(e: any): void {
+    this.selectMatch(this._active, e);
   }
 
   prevActiveMatch(): void {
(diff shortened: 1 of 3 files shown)


cakeinpanic:
@valorkin review please

src/typeahead/typeahead-container.component.ts
```diff
@@ -100,8 +100,8 @@ export class TypeaheadContainerComponent {
     return this.parent ? this.parent.typeaheadItemTemplate : undefined;
   }
 
-  selectActiveMatch(): void {
-    this.selectMatch(this._active);
+  selectActiveMatch(e: any): void {
```
valorkin, line 103: Should have a type
Name should be descriptive
Should be optional or will cause breaking change, and will have to wait for major release to piblish

src/typeahead/typeahead-container.component.ts
```diff
@@ -188,6 +188,9 @@ export class TypeaheadContainerComponent {
       e.preventDefault();
     }
     this.parent.changeModel(value);
+
+    value.selectEvent = e;
```
valorkin, line 192: I don't see usage of this field, plus it should be set to null on destroy.
Cuz event.target has link to DOM element which in turn could lead to memory leak via detached element persistance
In short: could be deleted?

valorkin (changes requested):

IraErshova:
Close this PR because of inactivity. Also, provide another solution for #4685 issue

Issue #3838, opened 2018-02-19
Fix for collapse Animations


bevbomb:
### Bug description or feature request:
I have put together an example of using the AnimationBuilder to handle the collapse animations. by not relying on the classes from the original collapse methods from bootstrap and just using the builder to handle the start, transition and complete i was able to reproduce the collapse animations.

### Plunker/StackBlitz that reproduces the issue:
StackBlitz: https://stackblitz.com/edit/ngx-bootstrap-collapse
Plunker: https://embed.plnkr.co/IdSJviSEK8X2jjxJNI0o/

### Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: Latest

Angular: 5.0.0

Bootstrap: 4.0.0

bevbomb:
@evilstiefel I have added a Plunker link as the StackBlitz is working fine for all of my browsers

duki994:
@bevbomb @evilstiefel 

You have a bit of a problem with this directive. It can (and will) fire before child views are initialized and checked (example div with table child with *ngFor directive). Initial height will not be full element height so directive will "cut" element view in half.

It is simply fixed by adding ngAfterViewChecked() hook :)

```
@Directive({
  selector: '[appBsCollapse]',
})
export class CollapseAnimationDirective implements AfterViewChecked {

 ...

  public isViewChecked: boolean = false;

 ..

  ngAfterViewChecked() {
    this.isViewChecked = true;
  }

...

  /* allows to manually toggle content visibility */
  public toggle(): void {
    if (this.isViewChecked) {
      if (this.isExpanded) {
        this.hide();
      } else {
        this.show();
      }
    }

....

}
```

bevbomb:
Hey @duki994 
Sorry for the late reply! You are completely right :) i probably wouldn't have picked this up until using it in a live situation, I'll add your update in now.

rodrigoic23:
Hello, any update on this?

ParisZX:
Hi, there is a pure css workaround to the collapse animation. I'm using something like this:
```
.collapse {
  /* setup transition however you like */
  -webkit-transition: max-height 0.5s ease;
  -moz-transition: max-height 0.5s ease;
  -o-transition: max-height 0.5s ease;
  -ms-transition: max-height 0.5s ease;
  transition: max-height 0.5s ease;
  
  /* override display, overflow and visibility */
  display: block!important;
  overflow: hidden!important;
  visibility: visible!important;
  
[...]

Issue #2473, opened 2017-08-22
fix(collapse): switching CSS property when use CollapseModule.


lazy-detourer:
I used CollapseModule like the following URL:

http://valor-software.com/ngx-bootstrap/#/collapse

And I confirmed that the 'CSS display Property' switches to 'block' and 'none'.
But in my case, it want to be another like 'inline-block' or 'table' when shown.
'none' is okay. It is exactly what I want.
But 'block' is not. How can I change to another value?

valorkin:
no easy way, but you can `#col="bs-collapse"` and `col.display="'inline-block'"`

Issue #801, opened 2016-07-29
feat(animations): add animations to all components


masiamj:
**Blocked by**: https://github.com/angular/angular/issues/9947

Are animations for the accordion, alerts, etc. on the roadmap for this project? Would be huge to further reflect fully-featured Bootstrap components.

jurienhamaker:
+1

qjnz:
Yes, could someone please advise the timeline for releasing the transition module?

qjnz:
Hi guys,  from this file \ng2-bootstrap\components\collapse\collapse.directive.ts,  you guys have implemented the animation transition but commented them all out because of the following ng issue.

however, the issue 5984 has been closed in June 2016, can you guys advise when the animation will be enabled for ng2-bootstrap please?

// fix: replace with // '@angular/animate';
// when https://github.com/angular/angular/issues/5984 will be fixed

raghavan85:
+1 Animations are quintessential for something like Accordion.

wolfgangmeyers:
Any updates on this?

valorkin:
ok, with angular2 2.1 animations became a thing, so they will be added soon

fareesh:
I needed accordion animations for a project so I made some very specific tweaks that work for my use-case. If anyone is interested you can check out: https://github.com/fareesh/ng2-bootstrap/commit/701af877bcc85f90eb21e6896cbdbfbff922393f

Would also be happy to make the necessary changes to make this usable in a generic case if someone can point me in the right direction.

Scipionh:
Hope the animations are coming soon, it's definitely a must.

zmnv:
Any solutions??

valorkin:
I can add almost all animations, but I will have to drop IE9 support
because they simply doesn't work there
plus at the moment it is hard to configure them

zmnv:
@valorkin да это какой-то беспредел вообще. юзаю ангуляр 2 юниверсал, там мне админ говорит мол используй версию ng2-bootstrap 1.1.16. Потому что jQuery там (в юниверсале) работает через задницу, он работает, но в компонентах другие плагины типа owl.carousel2 отрубаются. 

думал хоть это заработает. заработало. перелопатил css этот, зачем вообще использовать display, когда при разных значениях никакая анимация не работает. отрубаешь дисплей, меняешь на height: 0 (auto)  и т.д. и все работает. Уже третьи сутки сижу потею как черт))

[...]

PR #4814, opened 2018-11-16
Update of collapse directive

hc42:
Update of collapse directive:
- Added support for animations (pure bootstrap css)
- Removed non bootstrap classes
- Removed non bootsrap style modifications
- Added extra event emitters (animations introduce more states)
Unit Tests:
- Added bootstrap css to unit tests so that we test the "real (rendered) thing"
- Updated tests accordingly to new animations
Docs:
- Updated examples with animations
- Added new events docs
- Removed inline-block example (please look at the original bootstrap
css examples, this is really really not required and not the right way)

###
Targets there Issues
  #2473 
 #801 
#3838 

# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.


hc42:
Fix cypress test

valorkin:
Looks interesting, 1 thing is missing for sure:
Enabling animation should be done via config or/and input and disabled by default.
Or it will be a breaking change

Could you please add it?

hc42:
Finally found some time to do this. I added the input property "animate" (by default false). So this introduces no breaking change. However in the original JQuery-Bootstrap the default is enabled. It would be nice to have the same default behaviour some time in the future.

Domainv (approved):

karma-demo.conf.js
```diff
@@ -18,9 +18,10 @@ module.exports = function (config) {
       require('karma-coverage-istanbul-reporter'),
       require('@angular-devkit/build-angular/plugins/karma')
     ],
-    // files: [
+     files: [
     //   {pattern: './scripts/test.ts', watched: false}
-    // ],
+         "node_modules/bootstrap/dist/css/bootstrap.min.css",
```
ludmilanesvitiy, line 23: Why do you need this change?

src/spec/collapse.directive.spec.ts
```diff
@@ -43,45 +44,75 @@ describe('Directive: Collapse', () => {
...
+        detectCollapsed();
+      });
+    };
+    const triggerCollapse = () => {
+      detectCollapsing();
+      context.isCollapsed = true;
+      fixture.detectChanges();
+    };
+
+    triggerCollapse();
```
ludmilanesvitiy, line 83: in tests need to support AAA principle (arrange, act, assert)
I'm not sure why do you need triggerCollapse();

src/spec/typeahead-container.component.spec.ts
```diff
@@ -350,7 +362,8 @@ describe('Component: TypeaheadContainer', () => {
     });
 
     describe('prevActiveMatch', () => {
-      it('should select the last item and scroll to bottom', () => {
+      // broken test
+      xit('should select the last item and scroll to bottom', () => {
```
ludmilanesvitiy, line 366: we can't xit tests, need to fix/rewrite them

ludmilanesvitiy (changes requested):

karma-demo.conf.js
hc42: Short version: Without css available nothing would work.
Longer version: Animations are already defined by the css code. I try to not "reinvent the wheel" by not reimplementing bootstraps code in angular a second time. Actually I don't understand why it's often done in the project code base. Dosen't make sense to me.

src/spec/collapse.directive.spec.ts
hc42: Without calling this function the test would not be executed correctly. I grouped different steps  (points in time of the execution) in subfunctions because it seamed easier to read to me and it avoided callback nesting. Due to delay of animations it is required to wait for events and then check the state. One could also extend this test to for non expected events and throw an error in this case. However this arrangement seems to be simpler to read/understand (at least to me).

src/spec/typeahead-container.component.spec.ts
hc42: True, however this test fails in a complete different module and is not part of my changes. Sorry, but right now I don't have time to fix other bugs in other modules. You can leave the broken tests active of course.

Dassderdie:
Is there any news on this?

Domainv:
@Dassderdie The PR probably will be rejected
We decided to make more complex a general feature/refactor for animations of all components.

Dassderdie:
@Domainv Thx for the response :) 
Is there an ETA for the animations? It seems to me like there is a huge demand for them.

Domainv:
@Dassderdie For now no, general plans

Domainv:
Closes as resolved https://valor-software.com/ngx-bootstrap/#/collapse#animated

PR #4863, opened 2018-12-01
refactor(date-picker): remove redundant code

danielsogl:


src/datepicker/bs-daterangepicker-input.directive.ts
```diff
@@ -107,7 +107,7 @@ export class BsDaterangepickerInputDirective
     }
 
 
-    const _isDateValid = isDateValid(_value[0]) && isDateValid(_value[0]);
+    const _isDateValid = isDateValid(_value[0]);
```
valorkin: Hey, you are actually found a bug I suppose.
It should validate starting and ending dates in range.
And report invalid date

valorkin (changes requested):

src/datepicker/bs-daterangepicker-input.directive.ts
danielsogl: So I should change the validation like this?
```ts
const _isDateValid = isDateValid(_value[0]) && isDateValid(_value[1]);
```

valorkin: Rather _isDateValid0 and _isDateValid1,
So you could report invalid date, nearest if below

valorkin (approved):
Awesome, thanks!

PR #4957, opened 2018-12-19
fix(typeahead): fix typeahead's breaking if typeahead property receives NULL

ghost:
Closes to #4417

### DIFF
--- a/demo/src/app/components/+typeahead/demos/async/async.ts
+++ b/demo/src/app/components/+typeahead/demos/async/async.ts
@@ -78,9 +78,6 @@ export class DemoTypeaheadAsyncComponent {
 
   getStatesAsObservable(token: string): Observable<any> {
     const query = new RegExp(token, 'i');
-    if (!this.statesComplex) {
-      return of(null);
-    }
 
     return of(
       this.statesComplex.filter((state: any) => {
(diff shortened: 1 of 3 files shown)


src/typeahead/typeahead.directive.ts
```diff
@@ -448,6 +448,9 @@ export class TypeaheadDirective implements OnInit, OnDestroy {
   }
 
   protected finalizeAsyncCall(matches: TypeaheadMatch[]): void {
+    if (!matches) {
+      return;
```
valorkin:
```suggestion
       matches = [];
```

demo/src/app/components/+typeahead/demos/async/async.ts
```diff
@@ -78,6 +78,9 @@ export class DemoTypeaheadAsyncComponent {
 
   getStatesAsObservable(token: string): Observable<any> {
     const query = new RegExp(token, 'i');
+    if (!this.statesComplex) {
```
valorkin: remove after testing

valorkin (changes requested):

demo/src/app/components/+typeahead/demos/async/async.ts
ghost: removed

src/typeahead/typeahead.directive.ts
ghost: fixed in other way

valorkin (approved):

PR #4958, opened 2018-12-19
chore(build): use os specific path separator in npm run build.watch

mamohr:
The run build.watch task was not working in windows as it relied on the
task separator being '/'. Change it to use the OS specific separator to make it work in windows


ludmilanesvitiy:
Command `npm run build.watch` tested in development branch and in this branch. Tested in Linux, Windows7, windows10. In all those OSs works well. I'm not sure in what OS command did not work.

valorkin (approved):

valorkin:
Thank you )

PR #4992, opened 2019-01-08
Reset view state day to 1 when month or year is selected

### LINKED ISSUE #4396: Datepicker month bug
When I use ngx-bootstrap datepicker, the page just came in on May 31, I went to switch the month, clicked in June, but it came out in July.
(https://valor-software.com/ngx-bootstrap/old/2.0.4/#/datepicker)

ngx-bootstrap:2.0.2

Angular:4.0

Bootstrap:3

Angular CLI

johnsmathew:
When a month is selected, the 'day' field was not updated. This caused erroneous behavior in certain cases. For eg. if the current selected date is Oct 31, 2018 and user goes to the month selection view and chooses November, system tries to set the view state date to Nov 31, 2018. Since Nov has only 30 days, the current date is pushed to 1 Dec 2018. In effect, when Nov is chosen, the calendar for Dec was displayed.

When a month is selected, the view state day is now explicitly set to 1. Similarly, when an year is selected, month is set to 0 and day is set to 1.

Fixes #4396 

# PR Checklist

 - [x] added/updated tests.
 - [ ] added/updated demos.

### DIFF
--- a/src/datepicker/reducer/bs-datepicker.effects.ts
+++ b/src/datepicker/reducer/bs-datepicker.effects.ts
@@ -154,6 +154,7 @@ export class BsDatepickerEffects {
       this._store.dispatch(
         this._actions.navigateTo({
           unit: {
+            day: 1,
             month: getMonth(event.date),
             year: getFullYear(event.date)
           },
@@ -169,6 +170,8 @@ export class BsDatepickerEffects {
       this._store.dispatch(
         this._actions.navigateTo({
           unit: {
+            day: 1,
+            month: 0,
             year: getFullYear(event.date)
           },
           viewMode: 'month'
--- a/src/datepicker/bs-datepicker.spec.ts
+++ b/src/datepicker/bs-datepicker.spec.ts
@@ -49,8 +49,8 @@ describe('datepicker:', () => {
   let fixture: TestFixture;
   beforeEach(
     async(() => TestBed.configureTestingModule({
-        declarations: [TestComponent],
-        imports: [BsDatepickerModule.forRoot()]
+      declarations: [TestComponent],
+      imports: [BsDatepickerModule.forRoot()]
     }).compileComponents()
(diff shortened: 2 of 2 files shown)


src/datepicker/reducer/bs-datepicker.effects.ts
```diff
@@ -154,6 +154,7 @@ export class BsDatepickerEffects {
       this._store.dispatch(
         this._actions.navigateTo({
           unit: {
+            day: 1,
```
H--o-l, line 157: Thanks for this PR @johnsmathew !
Could something like "if selected day not in month, set it to day 1, else re-use the selected day" be possible ?
I think it would make a smoother UI.

adrienverge: I agree.

johnsmathew: From what I understood from the existing code, I don't think that is necessary. What we are passing to the `navigateTo` action is the view state. The calendar is displayed in such a way that the view state date is visible. So passing `day: 1` ensures that the correct month is displayed. Infact, passing any values from 1 to 28 would be same.

The previously selected date is stored in a separate variable and is unaffected. Until now, it was using the day from selected state for the view state as well. Maybe the PR title caused confusion. I should have named it as view state.

H--o-l: You are right @johnsmathew, I just tested it and I agree the behaviour is the same, expect for #4396 that is now fixed.
(I don't know if there is precedent in the project, but you could add a small comment at the end of the line, like `// dummy value so month do not overflow` or something like that).
Thanks for the fix, I look forward to see it in a release !

daniloff200:
That fix was applied in #5376 in a little bit other way, so, this one can be closed
cc @Domainv

PR #5005, opened 2019-01-15
Remove deprecated init event

FrancescoBorzi:
# PR Checklist

 - [x] added/updated tests.
 - [ ] added/updated API documentation. (no need)
 - [ ] added/updated demos. (no need)

The `Event.initEvent()` method has been deprecated and should no longer be used.

> This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

> Instead use specific event constructors, like Event().  ...

More info at: https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent


Domainv (commented):
@FrancescoBorzi 
please use cross-browser methods from library @netbasal/spectator
This library already added to the project and used in bs-datepicker.spec.ts

FrancescoBorzi:
@Domainv I don't know what that is and where you are referring to in the code.

However, isn't my code already cross-browser? It is the same way that Angular recommends:

https://angular.io/guide/testing#change-an-input-value-with-dispatchevent

Or maybe I misunderstood your comment?

FrancescoBorzi:
Also see: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent#Browser_Compatibility

Domainv:
@FrancescoBorzi 
IE11 doesn't support dispatchEvent
ie11_dsntsupport
screenshot from 2019-01-16 14-53-55

That's why we have to use cross-browser methods from library @netbasal/spectator to prevent the problems

FrancescoBorzi:
can you show me an example? I didn't find it in the code

Domainv:
@FrancescoBorzi 
In the PR I fixing the same problem
https://github.com/valor-software/ngx-bootstrap/pull/5009

Docs also
https://netbasal.gitbook.io/spectator/api

PR #5036, opened 2019-01-31
Bs datepicker custom classes

### LINKED ISSUE #2848: feat(datepicker): add customClass option to the new version of datepicker
Hi,
I'm using the legacy version of your datepicker and I'm heavily using the customClass attribute in order to set different classes on different dates according to the input data and I'm using it in inline mode.
I just updated to the latest version and now my legacy datepicker doesnt use momentjs locale as before and it goes back to the EN version.
I don't mind upgrading to the new version of the datepicker, but i cant seem to find a way to use customClass option over there.
Any advise will be highly appriciated.
Thanks.

### LINKED ISSUE #3958: feat(datepicker): Style/mark individual dates in datepicker
I need functionality in the datepicker to style individual dates, like in the old datepicker (customClass). I could submit a pull request if we first can agree of the approach to solve it.

One suggestion is to add an array of dates with according css class to DatepickerRenderOptions and then use it in BsDaysCalendarView. What do you think?

### LINKED ISSUE #4993: feat(datepicker): add possibility of keeping special dates always highlighted regardless of the selected date
The title already describes the feature. You've got a list of special dates that should be highlighted in the datepicker. This feature might be quite similar to #3192. I think the feature can be realized via the `BsDatepickerConfig`. Add here two attributes (something like that):

1.  `customSpecialDatesClass: string` - it expects a css class for highlighting the special dates
2. `specialDates: Date[]` it expects a list of dates that represents the dates that should be highlighted

martintro:
Closes #3958
Closes #4993 
Closes #2848 
# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.


Domainv (dismissed):

ludmilanesvitiy:
1) Need to change demo (for possibility have selected class green) 
2) Need to fix http://localhost:4200/#/datepicker#today-class functionality

ludmilanesvitiy:
Retested, looks good. Checked in Chrome, FF, IE.

PR #5085, opened 2019-03-01
fix(modal): added import strategy for lazy modules

thekiba:
BsModalModule is not supported lazy modules currently.

I made the separate method to explicit registration of lazy modules. This is not a solution to the problem.

This method should be able to add a new ComponentFactoryResolver from the current module to the ComponentLoaderFactory. We can discuss about that and implement it.

Related #4370

@valorkin 🗽


valorkin (approved):
thanks for contribution! not necessary change, but valid one

daniloff200 (approved):

PR #5099, opened 2019-03-11
fix(dropdown) fixing performance by removing @Hostlistnere('document:…

### LINKED ISSUE #2172: fix(dropdown): performance problems with multiple dropdowns

warash:
Closes #2172

@HostListener('document:click')  is causing huge performance issue. especially if there are many dropdown instances on page. Every instance is assigning global listener to click. Every listnere is causing appRef.tick being triggered. if you have 10 dropdowns on page and you click wherever on page 10 appRef.tick with full changedetection of all elements will be triggered.


Domainv:
@warash Thank you for the PR.
Now a little loaded will review it as soon as possible

ludmilanesvitiy:
Screenshot from 2019-03-26 15-11-07  ^^^current performance
Screenshot from 2019-03-26 15-06-10  
^^^performance with applying 4605 PR
Screenshot from 2019-03-26 15-05-51 - ^^^performance without any fixes
Also need to pay attention to runOutsideAngular function, which could time to time broke auto-waiting for dropdown open/close inside protractor tests, if they will exist in the app.

dklmuc:
imho there won't be any big performance improvements because of https://github.com/valor-software/ngx-bootstrap/pull/5099/files#diff-d74026377f7ce8e580379903df8e14beR78

warash:
hi @ludmilanesvitiy  i like also 4605 i would be super happy if either 4605 or mine pr would be merged into release.. it has big performance impact even bigger than on presented by you demo pages. becasue in real case scenario changedetection is much longer than on demo page that has no other components instead just dropdowns.. on our page we have pretty haevy kendo grid component which is changedetected multiple of times when clicked. Thanks for making this performance tests.

If it comes to runoutside angular and protractor problem i think it wont be a problem casue i am returing to angular zone by this.ngZone.run() when dropdown is realy opened and we need to inform angular about change.

ludmilanesvitiy:
@warash @dklmuc No worries, guys! This change request is under control and should be in the nearest version bump. I'll let you know!

Domainv:
Closes to prefer #4605

PR #5163, opened 2019-04-12
fix(datepicker): use ng-container to wrap z-w-s and button in month view

okonomiyaki3000:
Zero-width-space must be wrapped together with button or else, when the button is not shown, there will be an extra zero-width-space which will throw the remaining elements out of alignment.

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.

No, I didn't do any of those things. Nevertheless, this is the correct way.


valorkin (approved):

daniloff200 (approved):

dmitry-zhemchugov:
I have tested this PR. Now buttons are centered. Looks good

Issue #4853, opened 2018-11-29
Request: Reset interval carousel interval


marjoforcado:
I would like a feature that resets the interval of the carousel when you manually change the slides (by clicking on the indicator). Whats happening is that when you manually change slides the carousel keeps the current interval making the carousel switch to the next slide sooner than it has to be.

ludmilanesvitiy:
Closed by #5201, available from 5.0.0 version.

PR #5235, opened 2019-05-23
Fix/carousel reset time out #5201

### LINKED ISSUE #5201: Carousel: Reset time-out when clicking on arrow or the bottom indicators
I noticed that if you click one of the arrows of indicators, the timout for the next slide isn't reset. The problem with this is that if there is a time-out of 10 seconds, and you click the arrow after 9 seconds, you only see the this slide for 1 second and then it immediatly goes to the next slide.

Can you provide a fix for it?

gioboa:
Close: #5201 #4853

**Fixed behavior**
There is a time-out of 10 seconds.
After 9 seconds you select a new slide, the slide appear only for 1 second and then immediately goes to the next slide.


ludmilanesvitiy:
Tested interval demo and carousel at all in Chrome, IE - looks good.

PR #5326, opened 2019-07-12
fix(modal): change the mouse event to dismiss a modal on backdrop click

### LINKED ISSUE #5264: Modal closes on release of mouse outside modal box
I have built an example that will guide you through the odd behavior of modals:

https://stackblitz.com/edit/ngx-bootstrap-qdilhe?file=app/app.module.ts

To make it simple, If you click inside the modal, but drag and release the mouse outside the modal, the modal is dismissed nevertheless. This behavior only happens on `ngx-bootstrap`, not on bootstrap's modal.

florenthobein:
Because of the use of a 'click' event to trigger the dismissal of a modal, the target of the event doesn't take into account an eventual movement of the mouse before releasing click (for ex. selection of text), which can lead to unwanted dismissal of the modal.

Closes: #5264

# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated API documentation (not needed)
 - [x] added/updated demos (not needed)


src/modal/modal-container.component.ts
```diff
@@ -76,12 +77,18 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
     }
   }
 
-  @HostListener('click', ['$event'])
-  onClick(event: MouseEvent): void {
+  @HostListener('mousedown', ['$event'])
+  onClickStarted(event: MouseEvent): void {
+    this.clickStartedInside = event.target !== this._element.nativeElement;
```
benjaminrau: Property name or condition should be changed. Actually this evals to clickStartedOutside.
Changing condition would result in
`this.clickStartedInside = event.target === this._element.nativeElement;`

src/modal/modal-container.component.ts
```diff
@@ -76,12 +77,18 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
...
+  }
+
+  @HostListener('mouseup', ['$event'])
+  onClickStop(event: MouseEvent): void {
+    const clickedInside = this.clickStartedInside || event.target !== this._element.nativeElement;
```
benjaminrau: IMO should be
`const clickedOutside = !this.clickStartedInside || event.target !== this._element.nativeElement;`

src/modal/modal-container.component.ts
```diff
@@ -76,12 +77,18 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
...
     if (
       this.config.ignoreBackdropClick ||
       this.config.backdrop === 'static' ||
-      event.target !== this._element.nativeElement
+      clickedInside
```
benjaminrau: Should be changed then to use `clickedOutside `

src/modal/modal-container.component.ts
```diff
@@ -76,12 +77,18 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
...
     ) {
       return;
     }
```
benjaminrau, line 96: You could consider to reset`clickStartedInside` to default value after it was used for condition to not use a previous result of onClickStart if there was not mousedown inside but rather only mouseup.

benjaminrau (changes requested):
Thanks for the fix which is bugging us also. I tried to provide a review.

src/modal/modal-container.component.ts
```diff
@@ -38,6 +38,7 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
   isAnimated: boolean;
   bsModalService: BsModalService;
   private isModalHiding = false;
+  private clickStartedInside = true;
```
benjaminrau: The default value should be `false` because if there is no mousedown (started outside) the constraint is not checked if it started inside but value would be true.

benjaminrau (changes requested):

src/modal/modal-container.component.ts
florenthobein: You're right, there is a misunderstanding in the naming. `this._element.nativeElement` evaluates to the backdrop, that's why if `event.target` is different, it means we didn't click on the backdrop, hence we clicked on the modal. So a more precise name would be `clickStartedInsideModalContainer` (or `clickStartedOutsideBackdrop` but it feels like a double negative)

benjaminrau:
@florenthobein We tested your changes in our project and it didnt fix the issue in first place.
We use ModalDirective instead of the Component. After we applied your changes on the directive it is fixed in both.

I think the Directive should be fixed within this pull request also.

korinovsky (dismissed):

Lexikus:
How is this task going? It might fix a problem that I have.

florenthobein:
I just pushed an updated version taking in consideration comments about variable naming & including this fix in the context of a directive.

Sadly enough, there is no shared code between the `ModalDirective` and the `ModalContainerComponent`. I didn't take the risk to initiate a refactoring of this component, and had to duplicate part of the logic.

Tests are OK. Any update on a potential merge? 🙏

daniloff200:
Sorry for the long delay, I'm gonna review it and share with the QA team very soon

florenthobein:
There is something I would like to get your attention on, I have a strange behavior on my app using this patch and I'm not quite sure it's linked to this PR yet. When trying to scroll the modal using the mouse on scrollbar (drag&drop of the scrollbar), it closes the modal (verified on Chrome on a Macbook without mouse plugged). Do you reproduce this same behavior?

valorkin (approved):
thanks for contribution, looks good to me

daniloff200:
@florenthobein well, I hope @dmitry-zhemchugov  will help us to find out, is it an issue, related to this PR, or, something new, or, some special use case, which you found suddenly!

dmitry-zhemchugov:
via Hubstaff
User: Dmitry Zhemchugov

Project: int_ngx-bootstrap (OSS) - https://app.hubstaff.com/projects/304822
Date Range: 01/20/20 - 01/20/20
Work session total: 5:32:25
Billable: Yes

Grand total: 5:32:25

PR #5334, opened 2019-07-19
timepicker sets the dirty flag via writeValue

washtubs:
Basically the issue is that if you give the timepicker an initial value or you set it manually (as opposed to the user interacting with the element directly), it will set the dirty flag to true. This happens as a part of the `onChange` function. The dirty flag should stay false until a user interacts with the element. So only those functions should result in calling `onChange` i.e.: `changeHours`, `changeMinutes`, `changeSeconds`, `updateHours`, `updateMinutes`, `updateSeconds`, `toggleMeridian`. 

In general `writeValue` should never call `onChange`, which is essentially what was happening here.

By the way, my company is stuck on 3.x at this time. If this change is acceptable, is it also possible to get a patch for 3.x?

Demo on [stackblitz](https://stackblitz.com/edit/angular-ngx-bootstrap-timepicker-dirty).


daniloff200 (approved):

valorkin:
@washtubs hey, yes you can do PR to 3.x tag, and referer here
mentioning me will help to do it faster

src/timepicker/timepicker.component.ts
```diff
@@ -177,7 +178,6 @@ export class TimepickerComponent
       .subscribe((value: Date) => {
         // update UI values if date changed
         this._renderTime(value);
-        this.onChange(value);
```
valorkin, line 183: there are a chance that without mark for check it will not work

valorkin (commented):
@IraErshova could you please check my comment

daniloff200:
@washtubs  is it still actual change for you?

daniloff200:
Well, 11 months passed, and there's still no response. Also, almost 2 years since it has been opened. I think I can close it for now, but, you always can ping me to reopen it back.

PR #5364, opened 2019-08-20
fix(datepicker): fix timezone issue

### LINKED ISSUE #3609: DatePicker timezone problem
`[bsConfig]="{ dateInputFormat: 'YYYY-MM-DD' }"`

when this format is used, i may expect the same entry come back regardless of client. because we are selecting a date not datetime. But its considering it as datetime and puts timezone info to it.

So, for example, 2018-01-30 00:00:00 becomes 2018-01-29 if your timezone is +

There is no config option to set timezone or any thing that allows to retreive the input directly. 

If there is a way, i am not aware but per the source code, i dont see any option that lets this.

Domainv:
Closes #3609

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


src/datepicker/bs-datepicker-inline.component.ts
```diff
@@ -23,8 +23,16 @@ export class BsDatepickerInlineDirective implements OnInit, OnDestroy, OnChanges
     if (this._bsValue === value) {
       return;
     }
-    this._bsValue = value;
-    this.bsValueChange.emit(value);
+
+    let date = value;
+
+    if (date) {
+      const timezoneOffset = value.getTimezoneOffset() * 60000;
```
imben1109: Is it possible to have service to configure custom timezone offset?

src/datepicker/bs-datepicker.component.ts
```diff
@@ -69,8 +69,15 @@ export class BsDatepickerDirective implements OnInit, OnDestroy, OnChanges {
     if (this._bsValue === value) {
       return;
     }
-    this._bsValue = value;
-    this.bsValueChange.emit(value);
+    let date = value;
+
+    if (date) {
+      const timezoneOffset = value.getTimezoneOffset() * 60000;
```
imben1109: Is this possible to have a service to have a custom timezone offset?

imben1109 (commented):
Yes. It fix the UTC timezone offset issue. it would reset timezone offset to 0. 
But the could be possible situation that is would be a custom timezone offset. 
For example, date range picker in Hong Kong (UTC +8).

I would like to select a date from 27/08/2018 - 28/08/2018
the UTC Date Object should be 26/08/2018 16:00- 27/08/2018 16:00

ludmilanesvitiy:
Peek 2019-08-27 11-04
When the user manually set date, it converts to the previous day.

ludmilanesvitiy:
Same shift with min-mode
Screenshot from 2019-08-27 11-14-33

ludmilanesvitiy:
Smoke tested, looks good.

src/datepicker/bs-datepicker-inline.component.ts
Domainv: Maybe as different feature-request

src/datepicker/bs-datepicker.component.ts
Domainv: Maybe as different feature-request

PR #5385, opened 2019-09-10
fix(positioning): resolve perfomance issue

### LINKED ISSUE #5291: perf: PositionService creates active scroll & resize listener by default - even if not used
The PositionService creates a scroll & resize listener by default inside angular zone, regardless if you use it or not. 

https://github.com/valor-software/ngx-bootstrap/blob/8ce5e86dfcee4244bba897a163d31f71b795d8ba/src/positioning/positioning.service.ts#L58-L59

image

So just by loading ngx-bootstrap, you now have a application change detection cycle on every scroll and every resize event.

This results in a performance graph like this when you scroll:

image

Resulting in massive amount of cpu usage.

If you zoom into one spike, you will see the initial scroll event and zone.js calling `applicationRef.tick()` aka global change detection

image

Ideally what you want is:
* run the computation outsize angular zone via `ngZone.runOutsideAngular`
* only if a value changes jump back into the zone or distribute changes via observables

Additionally on touch devices you slow down the scroll behavior, as the scroll listener is not `{passive: true}`, thus the device waits for the listener to finish before scrolling animation can start, as the browser awaits a possible `event.preventDefault()`

see https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners

### LINKED ISSUE #5335: Modal performance (rerender on scroll)
When injecting a BsModalService into the constructor, a rerender occurs with each scroll.
https://stackblitz.com/edit/angular-g61vpl
ngx-bootstrap: ^5.1.0
Angular: 7.0.3
Bootstrap: 4.1.3

Domainv:
Closes #5291
Closes #5335
# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


ludmilanesvitiy:
Smoke tested, LG

PR #5448, opened 2019-10-08
testcommit

chawdamrunal:
''passada la ' +(m.hours() !== 1)' is always true because it evaluates to a non-empty string.

src/chronos/test/locale/ca.spec.ts (line : 208)

('passada la ' +(m.hours() !== 1)

This rule applies when a condition check has always the same result.

For example, it occurs when constant-valued variables are used or the condition is redundant with previously checked one. This rule also applies when only one case is executed due to the result of switch condition check being always the same.

It could be a debugging code, but at worst it is a programmer's mistake that completely misses block of code intended to be executed.


daniloff200:
Please, fork ths repo and create PRs from your fork

And also, it's not good to just spam with such pull requests

PR #5449, opened 2019-10-08
targetOffsets as any

chawdamrunal:
src/positioning/modifiers/arrow.ts (Line 25)
    (targetOffsets as any)[side] -=
Property '(targetOffsets as any)[side]' is used on both the left and right sides of this compound assignment. Consider using non-compound assignment instead.
A reference is used on both sides of compound assignment operator
This rule applies when a reference is used on both sides of compound assignment operator.

If there is a same reference on both sides of compound assignment operator, the compound operator applies to the reference itself.

It may be typo or intention. But it is recommended to use clearer and shorter code

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


daniloff200:
See no reason for one space PR

PR #5496, opened 2019-10-14
fix(datepicker): Maintain selected date when same date selected  (#5209)

### LINKED ISSUE #5209: DatePicker display 'value' not updated when same date is selected again
I want to display 'Today' string instead of the date, when current date is selected.
I am using value property to display either 'Today' if current date is selected or display date if any other date is selected other than current date.

When you select current date the first time, the value in the input box updates to 'Today' which is expected. However if you re-select current date again, it doesn't display 'Today' but instead just displays the date.  Please see the StackBlitz example below to reproduce the issue.

Expected: 
Datepicker input should update value when same date is selected again.

Using Plunkr, StackBlitz is the best way to show your issue. Issues without link to an example of reproduction might be closed.

You can use one of starter templates:

StackBlitz: https://stackblitz.com/edit/ngx-bootstrap-datepicker-mbae2d?file=app%2Fapp.component.ts

ngx-bootstrap:

Angular: 5

Bootstrap: 3.3.7

castriganoj:
Use getTime() for JS dateTime compare.

closes #5209 

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


castriganoj:
This does not exactly solve what the issue requests. Based on what was found in code and what the issue requests this update does solve the requester's issue of maintaining the 'today' string. When the Date value is set in the bs-datepickerDirective component there is a check for the same value. I believe the intention was to check if the same date was selected and return immediately if so. Otherwise continue and propagate the change to the other components with this.bsValueChange.emit(value). It compares 2 date objects which would always return false even if they are the same date. 

To summarize, this change may solve a separate but overlapping issue. Custom text inputs would require separate and additional changes. I look forward to your feedback. Please let me know how to proceed. 

Thanks much, Jim.

daniloff200 (approved):

daniloff200:
I think, it can be useful.

daniloff200:
hey, @castriganoj 
Thanks again for your contribution ! Could you please fill out this [form](https://docs.google.com/forms/d/1tHjwGbh8CIrLcOjcMOHoY7xiLNXpQWNXySwhJbwNrH8/edit) so we can send some swag to you? Thanks!

PR #5509, opened 2019-10-17
feat(modal): interception of closure for modal

florenthobein:
Offers a mechanism to prevent the closure of modals, by providing closure interception functions. When a modal is closed (whatever the way), the functions are triggered and return promises. If any promises is rejected, the modal will not close.

Supposed to validate this todo: https://github.com/valor-software/ngx-bootstrap/blob/development/src/modal/modal.directive.ts#L184

I have made a section in the demo to reproduce the use, and will do the e2e tests if the concept is approved.

# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.


src/modal/modal-options.class.ts
```diff
@@ -34,6 +34,10 @@ export class ModalOptions {
    * Modal data
    */
   initialState?: Object;
+  /**
+   * Modal data
```
valorkin: @omfgnuts need a better idea here ;)

florenthobein: Whoops, indeed, comment should be more like `Functions to intercept the closure`

src/modal/modal.directive.ts
```diff
@@ -34,6 +36,8 @@ export class ModalDirective implements OnDestroy, OnInit {
     return this._config;
   }
 
+  @Input() closeInterceptors: CloseInterceptorFn[];
```
valorkin: no docs

src/modal/modal.directive.ts
```diff
@@ -173,19 +177,31 @@ export class ModalDirective implements OnDestroy, OnInit {
     });
   }
 
-  /** Allows to manually close modal */
-  hide(event?: Event): void {
+  /** Check if we can close the modal */
+  tryHiding(event?: Event): void {
```
valorkin: this is a public API you should not rename it, because it makes a breaking change
instead, make `tryHiding`, `hide`, `private _hide` methods, so the old one still works as intended

src/modal/modal.helpers.ts
```diff
@@ -0,0 +1,16 @@
+import { CloseInterceptorFn } from './models';
+
+export const iterateOverInterceptors = (closeInterceptors: CloseInterceptorFn[],
```
valorkin: your usages doesn't use all function potential, where did you took it?

src/modal/models/index.ts
```diff
@@ -23,3 +23,5 @@ export interface DismissReasons {
   BACKRDOP: string;
   ESC: string;
 }
+
+export type CloseInterceptorFn = (data?: any) => Promise<any>;
```
valorkin: haven't seen data to be used at all, plus not a fan of `any` usage? may we remove it for now or be more specific about it?

valorkin (dismissed):
thank you!
overall looks great to me, but requires small changes

valorkin:
cc @daniloff200

src/modal/models/index.ts
florenthobein: Indeed, I anticipated a bit too much here, I'm going to remove what's not needed.

src/modal/modal.helpers.ts
florenthobein: I build this feature thinking that there would be a use for several interceptors that we could chain, using the result from the first one in the second one. Something like this :
```js
const closeInterceptors = [
      () => {
        if (confirm('Do you really want to close?')) {
          return Promise.resolve('Yes');
        } else {
          return Promise.resolve('No');
        }
      },
      (response) => {
        if (confirm(`You said ${response}, but are you really sure?`)) {
          return Promise.resolve();
        } else {
          return Promise.reject();
        }
      }
    ];
```
Now, if we remove the idea that there is a transmission of data, it doesn't seem useful to have several closeInterceptors.
What would you prefer? To have this chaining with the transmission of the result (and I make `data` other than `any` using generic types), or I simplify and just allow to provide only one closeInterceptor?

vvolodin:
Any timeline for this? What needs to be done? I'm willing to help in order to get this feature merged.

florenthobein (commented):
@valorkin I just pushed some updates on the PR, simplifying the parameter passed in ModalOptions (it is now a simple function).
I also took in account your previous review, and added a demo case and cypress testing.
For me it's good to go 👍 let me know if there is any other changes required.

daniloff200 (approved):

PR #5570, opened 2019-12-13
fix(datepicker):add reload to bsDatePickerInline on config change after init

mathias3d:
update bsDatePickerInline when config is set after init so the view reflects the latest restrictions

# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.


valorkin (approved):
this may not work in some circumstances, can't refer for sure
but had to add timeout once, in localizations demo
just to keep in mind

Issue #5576, opened 2019-12-20
Datepicker shows Invalid Date on picker and input field


arvindm95:
Bug description:
Random string entered on datepicker throws Invalid Date message on the Datepicker container/popover (Attached screenshot) and on the input field. 
<img width="352" alt="Screen Shot 2019-12-20 at 10 57 12 PM" src="https://user-images.githubusercontent.com/11587086/71293439-68994680-239b-11ea-9b54-1ed2a7edc4ce.png">

Plunker/StackBlitz that reproduces the issue:

https://stackblitz.com/edit/ngx-datepicker-bug

1. Open the above stackBlitz.
2. Type random string in input field.
3. Click outside of input field or click on any date.
4. The input field shows `Invalid Date` message.

Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: 5.3.2

Angular: 8.2.0

Bootstrap: 4.2.1

Build system: Angular CLI, System.js, webpack, starter seed:

**Expected behavior**

The input field should not display `Invalid Date` message. Rather it should do nothing. Just display the string entered by user as is. If a date is chosen from datepicker then the date overrides the value on the input field.

daniloff200:
hey, @arvindm95 !
I checked that PR, that works exactly the same, as you mentioned, but, i was able to reproduce that only for input, in DatePicker container it was not reproducible 

Thanks again for your contribution!

daniloff200:
cc @dmitry-zhemchugov  we need to discuss that with you later

valorkin:
closing for now, if you will help me to reproduce, I will help to fix

PR #5577, opened 2019-12-20
fix(datepicker): change to handle random string values on input field…

### LINKED ISSUE #5558: User is allowed to type Alphabets in DatePicker.
Bug description:  If you go to datePicker and start typing characters in it, it accepts it and then you select any date, its reflected as Invalid Date. 

ngx-bootstrap: 5.2.0

Angular: 8

Bootstrap: 4.3.1

**Expected behavior**

User can type numerical values but he should be restricted to type alphabets.

arvindm95:
closes #5558 
#5576

### DIFF
--- a/cypress/integration/datepicker_page_spec.ts
+++ b/cypress/integration/datepicker_page_spec.ts
@@ -155,10 +155,10 @@ describe('Datepicker demo page test suite', () => {
       datepicker.isDatepickerNavigationFullyActiveAndCorrect('date', 'body', datepicker.monthNames[4], '2015');
     });
 
-    it('when user clears input and add there date in bad format, click "Enter" - "Invalid date" shown', () => {
-      datepicker.clearInputAndSendKeys(basic, '20,10,2015', 0);
+    it('when user clears input and add there date in bad values, click "Enter" - Invalid Date is not shown shown', () => {
+      datepicker.clearInputAndSendKeys(basic, 'some value', 0);
       datepicker.clickEnterOnInput(basic);
-      datepicker.isInputValueEqual(basic, 'Invalid date', 0);
+      datepicker.isInputValueEqual(basic, 'some value', 0);
       datepicker.clickOnDatepickerInput(basic);
       datepicker.isSelectedDateExist('datepicker', false);
     });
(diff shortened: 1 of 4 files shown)


arvindm95:
fixed datepicker issue that throws invalid date when random values are entered. 
@daniloff200 can you please review this?

cypress/integration/datepicker_page_spec.ts
```diff
@@ -155,10 +155,10 @@ describe('Datepicker demo page test suite', () => {
       datepicker.isDatepickerNavigationFullyActiveAndCorrect('date', 'body', datepicker.monthNames[4], '2015');
     });
 
-    it('when user clears input and add there date in bad format, click "Enter" - "Invalid date" shown', () => {
-      datepicker.clearInputAndSendKeys(basic, '20,10,2015', 0);
```
valorkin, line 159: changes to this file doesn't seem to be appropriate, please revert and express what you are planning to test here

src/datepicker/bs-datepicker.component.ts
```diff
@@ -70,7 +71,10 @@ export class BsDatepickerDirective implements OnInit, OnDestroy, OnChanges {
       return;
     }
     this._bsValue = value;
-    this.bsValueChange.emit(value);
+
+    if (isValidDate(this._bsValue)) {
```
valorkin, line 75: why it's should not notify consumers about date change?

valorkin (changes requested):

src/datepicker/bs-datepicker.component.ts
arvindm95: Hi, In case of any invalid date entered (like alphabets, or any random string), the entire picker values show invalid date as below,
image

Hence, to avoid this, when an invalid date is typed, it is not sent further to consumers and stopped here.
#5576 - Please check this for reproduction and expected behavior.

cypress/integration/datepicker_page_spec.ts
arvindm95: Hi, I just modified the test case to work based on the fix to this issue #5576 . 
When invalid date is entered, it is retained and not passed on to reflect on the datepicker.

src/datepicker/bs-datepicker.component.ts
valorkin: sorry can't reproduce

PR #5616, opened 2020-01-23
fix(typeahead): fix broken typeahead property

daniloff200:
Rebased from commit after 5.4.0 release

Should fix broken typeahead property, which was found in #5589


valorkin (dismissed):

dmitry-zhemchugov:
Tested, Ready to merge. Working as expected.

PR #5645, opened 2020-02-05
feat(datepicker): add enabledDates functionality

### LINKED ISSUE #5121: Feature Request: Datepicker enabledDates
##Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: 3.3.0

Angular: 7.2.9

Bootstrap: 3.3.7

Build system: Angular CLI

----------------------------------------------------------------------

I have some cases where I would like to use the opposite of `disabledDates`, namely `enabledDates`.

i.e. with a property like

    datesEnabled = [ new Date('2019-03-21'), new Date('2018-03-21') ]

The only dates that should still be enabled in the datepicker like this, should then be `2019-03-21` and `2018-03-21`.

An example for this use case is a date picker that lets you only select dates for which you have any documents (i.e. online banking).

https://stackoverflow.com/questions/55276897/allowing-only-specific-dates-in-datepicker/55279232?noredirect=1#comment97324592_55279232

daniloff200:
resolves #5121 

# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.


src/datepicker/reducer/bs-datepicker.effects.ts
```diff
@@ -79,6 +79,12 @@ export class BsDatepickerEffects {
     return this;
   }
 
+  setDatesEnabled(value: Date[]) {
```
IraErshova: Could you please add type for returning data

src/datepicker/utils/bs-calendar-utils.ts
```diff
@@ -54,6 +54,14 @@ export function isDisabledDate(date: Date, datesDisabled: Date[]): boolean {
   return datesDisabled.some((dateDisabled: Date) => isSame(date, dateDisabled, 'date'));
 }
 
+export function isEnabledDate(date: Date, datesEnabled: Date[]): boolean {
+  if (datesEnabled === undefined || !datesEnabled || !datesEnabled.length) {
```
IraErshova: What's the difference between `datesEnabled === undefined` and `!datesEnabled` ?
Where is check that datesEnabled isArray?
Where is check that array consists only from Date instance? What if I pass only numbers or strings to this array?

src/datepicker/bs-daterangepicker-inline.component.ts
```diff
@@ -55,6 +55,10 @@ export class BsDaterangepickerInlineDirective implements OnInit, OnDestroy, OnCh
      * Disable specific dates
      */
     @Input() datesDisabled: Date[];
+    /**
+     * Disable specific dates
+     */
+    @Input() datesEnabled: Date[];
```
IraErshova, line 65: Why this property is mandatory for this directive? Is it correct?

src/datepicker/bs-daterangepicker.component.ts
```diff
@@ -110,6 +110,10 @@ export class BsDaterangepickerDirective
    * Disable specific dates
    */
   @Input() datesDisabled: Date[];
+  /**
+   * Enable specific dates
+   */
+  @Input() datesEnabled: Date[];
```
IraErshova, line 121: Why this property is mandatory for this directive? Is it correct?

src/datepicker/engine/flag-days.calendar.spec.ts
```diff
@@ -2,42 +2,85 @@ import { flagDaysCalendar } from './flag-days-calendar';
...
+            { date: new Date('2020-02-09'), label: '2020-02-09' }
+          ]
+        }
+      ],
+      weekNumbers: [],
+      weekdays: [],
+      monthTitle: '',
+      yearTitle: ''
+    };
+    const datesEnabled = [
```
IraErshova, line 63: I don't know where I should add this comment. Let it be here :) What will be if the user set up **datesEnabled** and **datesDisabled** at the same time? What will be if these dates are the same?

IraErshova (changes requested):
Could you please check my comment?

src/datepicker/bs-daterangepicker-inline.component.ts
daniloff200: for daterangepicker-inline and datepicker-inline all Inputs are mandatory.
Maybe, that was wrong from the beginning, and we should change that

src/datepicker/bs-daterangepicker.component.ts
daniloff200: replied above

src/datepicker/engine/flag-days.calendar.spec.ts
daniloff200: :thinking: 
I don't expect somebody will use both of this functionalities, that's really strange
If dates are the same, all dates will be disabled at all

src/datepicker/utils/bs-calendar-utils.ts
daniloff200: if you pass something else, which is not a Date instance, that will not successfully pass isSame check, so, that nothing will be enabled / or disabled, because we have exactly the same function for datesDisabled 

About typecheck -> will be reworked right now

src/datepicker/engine/flag-days.calendar.spec.ts
daniloff200: added a note for both functionalities in doc, about not using both of them at the same time

IraErshova (dismissed):

IraErshova (dismissed):

dmitry-zhemchugov:
Tested. Ready to merge

IraErshova (approved):

PR #5861, opened 2020-07-16
fix(modal): remove the delay before hiding the modal

### LINKED ISSUE #1938: Modal delay on closing
What can be issue?
In demo i see normal modal, but when included in my angular 2 project, on modal close it delays for ~100ms?

### LINKED ISSUE #3711: Modal created via ModalService is not hidden after calling .hide()
Calling the hide() function on a modal will not hide it if it's not shown fully. This means that if I open a dialog to signal a usually long loading server call which finishes too fast (faster than the modal animation) the hide() will be called but has no effect. You can see this in my demo with three manual delays. At no or small delays the dialog just stays open, at 150ms it isn't even shown properly and at 250ms it works as expected. Another interesting behaviour is that opening many dialogs and closing them fast will leave the GUI unclickable (E.g. in my demo by clicking and holding ctrl+2).

StackBlitz: https://stackblitz.com/edit/ngx-bootstrap-ftrzc4?file=app%2Fapp.component.ts

ngx-bootstrap: 2.0.2

Angular: 4.4.6

Bootstrap: 3.3.7

Meigyoku-Thmn:
# PR Checklist

 - [ ] added/updated tests.
 - [ ] added/updated demos.

<[These lines](https://github.com/valor-software/ngx-bootstrap/blob/539dc28e4584754f2a0feee3073353f6159885c2/src/modal/bs-modal.service.ts#L140)> aren't making sense, why `bsModalRef.hide()` has to wait for a bit before doing the real hide routine?  I consider it as a bug (there is no mention about this in the docs).
I saw someone has made a change: https://github.com/valor-software/ngx-bootstrap/pull/4550 to fix https://github.com/valor-software/ngx-bootstrap/issues/3711, yes, that fixed their problem, but causes new problem for anybody else. Honestly, if you guy want to slow down a little bit before closing modal, just setTimeout for yourself.
This will resolve https://github.com/valor-software/ngx-bootstrap/issues/1938.


pavankjadda:
You can set `animation: false` to fix this

```
   modalConfig = {
        animated: false,
        keyboard: false,
        backdrop: true,
        ignoreBackdropClick: true,
        class: 'modal-lg'
    };

```

Meigyoku-Thmn:
But that config disables every animations, not just the delay. And no, I don't want to make animation css rule myself (can't even control the right html element without making a global css rule which I opt to avoid).

pavankjadda:
I agree, it was temporary work around. This PR should fix the issue

daniloff200:
If you ask me, I also don't like the previous solution, but, it was important to fix those issue.
However, the delay is really annoying, so, I think, we can revert this back

daniloff200 (dismissed):

daniloff200 (approved):

PR #5870, opened 2020-07-22
refactor(configs): make component configs be tree-shakable

### LINKED ISSUE #5824: Add provideIn: 'root' to config services
Almost all the component modules register providers using 'forRoot' strategy, this makes the code not treeshakeable because all the package is included in main.js instead of the lazy route really using it.

If all the global config services has the @Injectable({ providedIn: 'root' }) decorator there is no need of the module "forRoot" method in some cases as the ProgressBarModule, and the tree shake will create a chunk with the package reference instead of add it in main.js.

Eg:
```typescript
@Injectable()
export class ProgressbarConfig {
  /** if `true` changing value of progress bar will be animated */
  animate: Boolean = false;
  /** maximum total value of progress element */
  max = 100;
}
```
Change to:
```typescript
@Injectable({
  providedIn: 'root'
})
export class ProgressbarConfig {
  /** if `true` changing value of progress bar will be animated */
  animate: Boolean = false;
  /** maximum total value of progress element */
  max = 100;
}
```

ilyaryabchinski:
add `providedIn: root` property to all Injectable decorators of component config services

closes #5824


valorkin:
Thank you for a PR. It introduces a breaking change to API, let's schedule it for the v7 release. 
cc @daniloff200

valorkin (approved):
looks good to me

PR #5878, opened 2020-07-29
increase z-index of bs-dropdown-container

woricek:
In newest version of bootstrap, modal uses z-index 1050 and dropdown container was hidden under modal. With z-index: 1050 it works fine.


daniloff200:
Well, I made this change in January, I suppose.
@woricek could you, please, provide an example, or, demo, how it looks like, and why exactly this change needed ?
Or, at least a way how to reproduce it
Thanks in advance!

daniloff200:
Closes because of inactivity and no response from the author

Feel free to write some comment here, and ping me, and I'll reopen this PR

PR #5940, opened 2020-09-29
fix(modal): typo bug in test code (modal.service.spec)

leo6104:
# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.

Thank you for accepting #5939 PR. 
I found a typo-error in my PR but it already accepted to merge. 
So i create new PR to fix the test code. Sorry for my mistake. 😅


daniloff200:
That's really strange for me, I'm not really sure, that it was an error in tests, because, in that case they would be failed, but, they all passed :smiley: 
But yeah, maybe, in that case it was just 'undefined' value, so, it worked in 100% cases, even when it should not

daniloff200 (approved):

PR #5941, opened 2020-09-29
feat(Modal): enable type-checking for BsModalRef.content & ModalOptions.initialState

### LINKED ISSUE #4561: ModalOptions#initialState - better type definition
As I have found in one of the issues, `initialState` property of ModalOptions should be used to pass arguments to modal constructor.
But as long as the initialState has type Object, those properties cannot be read easily (e.g. I have to do type assertion `(options.initialState as any).myProp`).
I would suggest to change `Object` to indexable type like: `{ [key: string]: any }`  

(https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types)

### LINKED ISSUE #5807: Make BsModalRef.content typesafe
**Is your feature request related to a problem? Please describe.**

I am using a lot of modals where I have to provide some objects that I will edit in modal.
Only way to do this is 
```ts
const modal: BsModalRef = this.bsModalService.show(MyModalComponent);
modal.content.object1 = this.object1;
modal.content.object2 = this.object2;
```

or to have some methods inside modal component
```ts
const modal: BsModalRef = this.bsModalService.show(MyModalComponent);
modal.content.setData(this.object1, this.object2);
```

Problem with both of those 2 implementations is that they are not typesafe, because `BsModalRef.content` is type of `any`.
Then I can get into trouble if I rename `MyModalComponent.object1` into `obj`, it won't be changed in other classes and there won't be any build error

**Describe the solution you'd like**

What would be nice to have is that `BsModalRef` also accepts some type, and then, content also get that type, and we have proper auto-completition and proper refactoring 

so, it should be something like this
```ts
export declare class BsModalRef<T> {
    /**
     * Reference to a component inside the modal. Null if modal's been created with TemplateRef
     */
    content?: T | null;
}
```
This would be used then easily
```ts
const myModalRef: BsModalRef<MyModalComponent> = this.bsModalService.show(MyModalComponent);
myModalRef.content.object1 = this.object1;
```
It will have with proper typing, and proper refactoring I would say

**Describe alternatives you've considered**

I have tested this solution locally, and it was working properly.
Refactoring is working, it is typesafe (when I remove one param from `modal.content.setData(this.object1, this.object2);` it is immediately underlined red in editor and I know I have some problem.

**Additional context**

leo6104:
Resolve #5807 #3504 
Close #5019
closes #4561 

# PR Checklist

 - [x] added/updated tests.
 - [x] added/updated demos.

```typescript
@Component({ ... })
export class TestComponent {
   firstProp: string;
   secondProp: number;
}

const ref = this.modalService.show(TestComponent, {
  initialState: { firstProp: 'test', secondProp: 1 } // type: Partial<TestComponent>
});
ref.content // type: TestComponent
```

### DIFF
--- a/src/modal/bs-modal-ref.service.ts
+++ b/src/modal/bs-modal-ref.service.ts
@@ -1,7 +1,8 @@
 import { Injectable, EventEmitter } from '@angular/core';
 
 @Injectable()
-export class BsModalRef {
+// tslint:disable-next-line:no-any
+export class BsModalRef<T = any> {
   /**
    *  Allow user to ID for the modal. Otherwise, a unique number will be given
    */
@@ -10,8 +11,7 @@ export class BsModalRef {
   /**
    * Reference to a component inside the modal. Null if modal's been created with TemplateRef
    */
-  // tslint:disable-next-line:no-any
-  content?: any | null;
+  content?: T | null;
 
   /**
    * Hides the modal
--- a/src/modal/bs-modal.service.ts
+++ b/src/modal/bs-modal.service.ts
@@ -65,8 +65,11 @@ export class BsModalService {
   }
 
   /** Shows a modal */
-  // tslint:disable-next-line:no-any
-  show(content: string | TemplateRef<any> | any, config?: ModalOptions): BsModalRef {
+  show<T = Object>(
+    // tslint:disable-next-line:no-any
+    content: string | TemplateRef<any> | { new(...args: any[]): T },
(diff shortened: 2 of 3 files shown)


daniloff200 (approved):

NenadJovicic:
Thanks for this
This is so useful 
In which version it will come?

daniloff200:
Well, I assume we gonna release a new version soon, with some small fixes and updates, hope it will be in next week

PR #5962, opened 2020-10-19
fix(progressbar): add aria-valuemax fix

### LINKED ISSUE #5957: NGX-Bootstrap Progressbar Component not updating aria-valuemax attribute
**Bug description:**

The **ProgressbarComponent** is not getting updated with the **aria-valuemax** attribute. Other aria attributes such **aria-valuenow** and **aria-valuetext** are getting updated whenever the **currentValue** or the **maxValue** of the **Progressbar** changes.
But the **aria-valuemax** attribute is not getting appended to the element even on load of the component.

This issue is reproducible in the **NGX-Bootstrap's** demo site:

https://valor-software.com/ngx-bootstrap/#/progressbar

**aria-valuemax** attribute not present in the below **Progressbar** example from **NGX-Bootstrap** Demo site:
image

ngx-bootstrap: v5.6.1

Angular: v9.1.4

Bootstrap: v3.4.1

**Expected behavior**

The **aria-valuemax** attribute should get appended to the **BarComponent** and also it should get updated whenever the **maxValue** changes.

**Root Cause**

The issue is with the "max" property of the **BarComponent** which is in-turn binded to the **aria-valuemax** attribute.
But this **"max"** property is neither an "@Input" like the **"value"** nor is updated whenever the **maxValue** changes.
It always remains undefined.

image

**Solution**

The **"max"** should be an "@Input" property so that it can be passed to the **BarComponent** from the **ProgressbarComponent** and so the **aria-valuemax** attribute will get updated whenever the **maxValue** changes.

daniloff200:
# PR Checklist

fixes #5957

 - [ ] added/updated tests.
 - [ ] added/updated demos.

### DIFF
--- a/src/progressbar/bar.component.ts
+++ b/src/progressbar/bar.component.ts
@@ -6,7 +6,7 @@ import {
   OnDestroy,
   OnInit,
   ElementRef,
-  Renderer2
+  Renderer2, SimpleChanges
 } from '@angular/core';
 
 import { ProgressbarComponent } from './progressbar.component';
@@ -30,32 +30,12 @@ import { isBs3 } from 'ngx-bootstrap/utils';
   }
 })
 export class BarComponent implements OnInit, OnDestroy {
-  max: number;
-
+  @Input() max: number;
   /** provide one of the four supported contextual classes: `success`, `info`, `warning`, `danger` */
-  @Input()
-  get type(): string {
-    return this._type;
-  }
-
-  set type(v: string) {
-    this._type = v;
-    this.applyTypeClasses();
-  }
+  @Input() type: string;
 
   /** current value of progress bar */
-  @Input()
-  get value(): number {
-    return this._value;
-  }
-
-  set value(v: number) {
-    if (!v && v !== 0) {
-      return;
-    }
-    this._value = v;
-    this.recalculatePercentage();
-  }
+  @Input() value: number;
 
   @HostBinding('style.width.%')
   get setBarWidth() {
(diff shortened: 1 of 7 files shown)


src/progressbar/bar.component.ts
```diff
@@ -30,7 +30,15 @@ import { isBs3 } from 'ngx-bootstrap/utils';
...
 export class BarComponent implements OnInit, OnDestroy {
-  max: number;
+  @Input()
+  get max(): number {
+    return this._max;
+  }
+
+  set max(v: number) {
+    this._max = v;
+    this.applyTypeClasses();
```
valorkin: please move this to `OnChange ` and react only on type\max changes

valorkin (changes requested):
one small change please

src/progressbar/bar.component.ts
daniloff200: done

valorkin (approved):
Looks good to me

src/progressbar/progressbar.component.html
```diff
@@ -1,6 +1,6 @@
-<bar [type]="type" [value]="_value" *ngIf="!isStacked">
+<bar [type]="type" [value]="_value" [max]="_max" *ngIf="!isStacked">
   <ng-content></ng-content>
 </bar>
 <ng-template [ngIf]="isStacked">
-  <bar *ngFor="let item of _value" [type]="item.type" [value]="item.value">{{ item.label }}</bar>
+  <bar *ngFor="let item of _value" [type]="item.type" [value]="item.value" [max]="item.max">{{ item.label }}</bar>
```
Sebi11, line 5: Hello!
I'm afraid this fix has broken the `max` attribute when using a "stacked" progress bar. This small piece of code requires each item of `BarValue` to have the same `max` attribute, rather than using the `max` input of the `ProgressbarComponent`.

Issue #5745, opened 2020-04-02
Issue while using dropdown(ngx-bootstrap) inside an accordian header.


experionakash:
Bug description:
I have a scenario like need to add a dropdown menu inside accordian header.

Current behaviour :
While i'm clicking on the dropdown in the accordian header both the accordian and the dropdown opens.

Plunker/StackBlitz that reproduces the issue:
https://stackblitz.com/edit/angular-x1u12g

ngx-bootstrap: 5.6.1

Angular: 8.0.0

Bootstrap: 

Build system: Angular CLI, 

**Expected behavior**
It should only open the dropdown.

I have tried to add the `event.stopPropagation()` and it woks, but then the issue is, dropdown wont close when i'm opening another dropdown.

PR #6154, opened 2021-05-28
feat(accordion): using dropdown(ngx-bootstrap) inside an accordion

SvetlanaMuravlova:
Issue while using dropdown(ngx-bootstrap) inside an accordion header.
refs #5745

### DIFF
--- a/src/accordion/accordion-group.component.ts
+++ b/src/accordion/accordion-group.component.ts
@@ -76,9 +76,14 @@ export class AccordionPanelComponent implements OnInit, OnDestroy {
     this.accordion.removeGroup(this);
   }
 
-  toggleOpen(): void {
+  toggleOpen(event: Event): void {
+    if (this.checkAccordionPanelToggle(event.target as HTMLElement)) return;
     if (!this.isDisabled) {
       this.isOpen = !this.isOpen;
     }
+  }
+
+  checkAccordionPanelToggle(target: HTMLElement): boolean {
+    return target && target.classList.contains('avoid-toggle')
   }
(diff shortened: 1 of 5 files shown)


src/accordion/accordion-group.component.ts
```diff
@@ -76,9 +76,14 @@ export class AccordionPanelComponent implements OnInit, OnDestroy {
     this.accordion.removeGroup(this);
   }
 
-  toggleOpen(): void {
+  toggleOpen(event: Event): void {
```
valorkin: This is public API, adding mandatory option is potentially a breaking change. Please make it optional

src/accordion/accordion-group.component.ts
```diff
@@ -76,9 +76,14 @@ export class AccordionPanelComponent implements OnInit, OnDestroy {
...
+    if (this.checkAccordionPanelToggle(event.target as HTMLElement)) return;
     if (!this.isDisabled) {
       this.isOpen = !this.isOpen;
     }
   }
+
+  checkAccordionPanelToggle(target: HTMLElement): boolean {
```
valorkin: This is static method, don't add it to class and enable 
Or make it static or extract it to helper file

valorkin (changes requested):
Please address comments

valorkin (changes requested):
should be resolved in dropdown toggle via https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

valorkin (approved):

PR #6156, opened 2021-05-28
Fix(datepicker):  ranges and  max date

### LINKED ISSUE #6105: maxDate and maxDateRange not working properly when used together in ngx-bootstrap date range picker
Hi 

I have set my maxDate to currentDate(today) and maxDateRange set to 10 days. So when I click on yesterday it still allows me to select future dates because of the range 10 set. But here, I have also set maxDate to today(currentdate). 
So when I using both together maxDate is not working when the range falls within.

https://stackblitz.com/edit/angular-djpvky?file=app/ngx-bootstrap-demo.component.ts

Use the above app to replicate the issue.
So by default future dates will be disabled, but after selecting current date future dates automatically enabled for the range mentioned but this should not be the case. It should work with max date mentioned.

Thanks
Vijai

Before opening an issue, please search for duplicates (opened and closed) https://github.com/valor-software/ngx-bootstrap/issues There's no need to open an issue here if you want to ask general question, use StackOverflow or Slack instead

Bug description:

Plunker/StackBlitz that reproduces the issue:

Using Plunkr, StackBlitz is the best way to show your issue. Issues without link to an example of reproduction might be closed.

You can use one of starter templates:

Plunkr: https://plnkr.co/edit/0NipkZrnckZZROAcnjzB?p=preview

StackBlitz: https://stackblitz.com/edit/ngx-bootstrap?file=app%2Fapp.module.ts

ngx-bootstrap:

Angular:

Bootstrap:

**Expected behavior**

SvetlanaMuravlova:
maxDate and maxDateRange not working properly when used together in ngx-bootstrap date range picker
refs #6105


daniloff200 (dismissed):

daniloff200 (dismissed):

valorkin (changes requested):
Missing semicolons

src/datepicker/bs-daterangepicker.component.ts
```diff
@@ -251,12 +252,46 @@ export class BsDaterangepickerDirective
...
   }
 
+  /**
+   * Checks max date and dates in ranges
+   * if one of dates in ranges is later then maxDate
+   * this date becomes equal to maxDate
+   */
+
+  checkMaxRangeDate(ranges?: BsCustomDates[], maxDate?: Date): BsCustomDates[] | void {
+    if (!ranges || !maxDate || (ranges && !ranges.length && !ranges[0].value)) return ranges;
```
valorkin: Please extract this method to helper file and cover with tests.
And don't nest ifs, prefer guard checks and early exits

valorkin (changes requested):
Please address comments

valorkin:
+ apply to other rangepickers as well

valorkin (approved):

Issue #6174, opened 2021-06-11
Ngx-bootstrap time picker wheel change event and meridian change event not found.


Sushma-tiwari:
Ngx-bootstrap time picker wheel change event and meridian change event not found.
Because i want validation like start and end time in which i am using two ngx-bootsrap time picker if user changes the value of meridian(clicking on am/pm button) then how can validate the thing?
![timepicker](https://user-images.githubusercontent.com/66738000/121630575-07879600-ca9b-11eb-824c-002b4b35741d.png)

PR #6256, opened 2021-08-04
feat(timepicker): added meredianChange output

SvetlanaMuravlova:
feat(timepicker): added changedMeridianValue output

refs #6174


src/timepicker/timepicker.component.ts
```diff
@@ -118,7 +118,7 @@ export class TimepickerComponent
 
   /** emits true if value is a valid date */
   @Output() isValid = new EventEmitter<boolean>();
-
+  @Output() changedMeridianValue = new EventEmitter<string>();
```
valorkin: I would use naming like here https://angular.io/guide/two-way-binding
`meredianChange`

src/timepicker/timepicker.component.ts
```diff
@@ -118,7 +118,7 @@ export class TimepickerComponent
 
   /** emits true if value is a valid date */
   @Output() isValid = new EventEmitter<boolean>();
-
+  @Output() meridianChange = new EventEmitter<string>();
```
valorkin, line 122: please add comment for a documentation

valorkin (changes requested):

valorkin (approved):

PR #6271, opened 2021-08-26
fix(popover, tooltip): fix positioning by not accessing always undefi…

### LINKED ISSUE #6237: Incorrect tooltip and popover position for inline elements after updating to 7.0.0
**Bug description:**

Tooltip appears at the top left of inline element like &lt;a&gt; and &lt;span&gt; instead of the middle. 
The position is correct if downgrade to ngx-bootstrap 6.2.0

**StackBlitz that reproduces the issue:**

https://stackblitz.com/edit/angular-obazeg

**Versions of ngx-bootstrap, Angular, and Bootstrap:**

ngx-bootstrap: 7.0.0

Angular: 12.1.0

Bootstrap: 4.5.3

Stubenhocker1399:
…ned values

The defined offsets result object never has right and bottom defined,
which led to the width and height values always returning 0 when sizes width or clientWidth
were undefined.
Before commit 51051bd6e0b9c9e9ecb05b271e4ec08fd9970358 this line calculated
'NaN' which caused the following scrollbar adjustment to not happen,
whereas now it subtracted the entire element width and height leading to
incorrect positioning of the popover and tooltips.

Closes #6237

# PR Checklist

 - [x] added/updated tests.


valorkin:
thanks :)

PR #6272, opened 2021-08-27
fixies(positioning): fixed positioning for inline elements

SvetlanaMuravlova:
fixies(positioning): fixed tooltip and popover positioning for inline elements


Stubenhocker1399:
This doesn't solve the core issue of this bug, see #6271.
`result.bottom` and `result.right` will always be undefined.

SvetlanaMuravlova:
such issues were resolved in other prs

Issue #6260, opened 2021-08-10
If datetiepicker used along with ngModel it replaces the initial input date with current date.


mukuljai:
@daniloff200 this change https://github.com/valor-software/ngx-bootstrap/pull/5701 is actually causing issue where if datetiepicker used along with ngModel it replaces the initial input date with current date. To pin point the issue in code here -
```
if (!this._bsValue && value) {
      const now = new Date();
```
Why need to have current date fixed in lib, why it cant be provided as input if required?

_Originally posted by @mukuljai in https://github.com/valor-software/ngx-bootstrap/issues/5701#issuecomment-895272330_

nielswitte:
I am facing a similar issue where multiple datepicker are used and after changing one date, alle the hours, minutes, seconds of the other dates are set to `now`  (the moment of initializing the datepicker). All Date objects have midnight UTC as time (hours, minutes, seconds all 0) on initializing the form, but after one change, alle hours, minutes and seconds are replaced.

PR #6288, opened 2021-09-10
fix(datepicker): added optional time initing

### LINKED ISSUE #6276: Datepicker modifies initial value after update from 6.2.0 to 7.0.0
**Bug description:**
After updating from 6.2.0 to 7.0.0, the datepicker changes the time of any `Date` value that is set to the users current local time. This also happens when the datepicker is used in a form and an initial value is set. This causes the initial form value to change, while the user has not made any changes. Also, setting the time of the `Date` value to the current time makes no sense, since the time of selection is not in any way relevant to the selected date.

The commit that causes this: https://github.com/valor-software/ngx-bootstrap/commit/bc188d69bf17010dadd924673bfa13eef3944833

I suggest that this behaviour be removed again. Alternatively, it might make more sense to set the time to start of day on the selected date.

**Versions of ngx-bootstrap, Angular, and Bootstrap:**

ngx-bootstrap: 7.0.0

Angular: 12.1.4

Bootstrap: 4.5.3

**Expected behavior**

Expected behaviour would be that the initial `Date` value is set without modifications.

SvetlanaMuravlova:
fix(datepicker): added optional time initing
closes #6276  #6260


daniloff200 (dismissed):

src/datepicker/bs-datepicker-inline.component.ts
```diff
@@ -31,7 +31,13 @@ export class BsDatepickerInlineDirective implements OnInit, OnDestroy, OnChanges
   /**
    * Config object for datepicker
    */
-  @Input() bsConfig?: Partial<BsDatepickerInlineConfig>;
+  @Input() set bsConfig(value: Partial<BsDatepickerInlineConfig>) {
```
valorkin: move to ngOnChanges

src/datepicker/bs-datepicker-inline.component.ts
```diff
@@ -179,15 +181,16 @@ export class BsDatepickerInlineDirective implements OnInit, OnDestroy, OnChanges
       this._datepicker.hide();
     }
 
-    this._config = Object.assign({}, this._config, this.bsConfig, {
-      value: checkBsValue(this._bsValue, this.maxDate || this.bsConfig && this.bsConfig.maxDate),
+    this._config = Object.assign({}, this._config, this._bsConfig, {
```
valorkin: revert

valorkin (changes requested):
please apply comments

valorkin (dismissed):

PR #6292, opened 2021-09-13
fix(modals): fixed modals documentation with adding typo ModalOptions

### LINKED ISSUE #6059: Upgrading 5->6 breaks modal creation with initialState
Upgrading from ngx-bootstrap 5 to 6 is causing a compiler error with my old code that sent initialState to a modal with a component. This used to work, and looks pretty much like the example in the docs:

    const initialState = {
        clientID: 123,
        username: 'foo',
    };
    this.modalRef = this.modalService.show(MyComponent, {initialState});

The error:
**Type '{ clientID: number; username: any; }' is not assignable to type 'Partial<OneTimePasscodeComponent>'.
  Types of property 'clientID' are incompatible.
    Type 'number' is not assignable to type 'string'.ts(2322)
modal-options.class.d.ts(38, 5): The expected type comes from property 'initialState' which is declared here on type 'ModalOptions<OneTimePasscodeComponent>'**

I can't create a Plunker or StackBlitz as both fail to install version 6.2.0 of ngx-bootstrap, the templates are for old versions and both sandboxes cannot seem to install v6.

I'm using Angular 9, bootstrap 4, ngx-bootstrap 6.2.0

SvetlanaMuravlova:
fix(modals): fixed modals documentation with adding typo ModalOptions
close #6059

### DIFF
--- a/libs/doc-pages/modal/src/lib/demos/service-component/service-component.ts
+++ b/libs/doc-pages/modal/src/lib/demos/service-component/service-component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { BsModalService, BsModalRef } from 'ngx-bootstrap/modal';
+import { BsModalService, BsModalRef, ModalOptions } from 'ngx-bootstrap/modal';
 
 @Component({
   // eslint-disable-next-line @angular-eslint/component-selector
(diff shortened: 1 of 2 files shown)


libs/doc-pages/modal/src/lib/demos/service-component/service-component.ts
```diff
@@ -20,7 +20,7 @@ export class DemoModalServiceFromComponent {
       ],
       title: 'Modal with component'
     };
-    this.bsModalRef = this.modalService.show(ModalContentComponent, {initialState});
+    this.bsModalRef = this.modalService.show(ModalContentComponent, { initialState } as ModalOptions);
```
valorkin: ` const initialState: ModalOptions = {` it's better to define at object init

valorkin (changes requested):

valorkin (dismissed):

PR #6318, opened 2021-10-01
chore(bump): added enableivy partial mode

SvetlanaMuravlova:
chore(bump): added enableivy partial mode


valorkin:
partial compilation enabled

PR #6339, opened 2021-10-29
fix(datepicker): fixed inline datepicker subscriptions

### LINKED ISSUE #6263: Setting datesDisabled breaks inline datepicker bsValueChange-Event
**Bug description**:

When using `bs-datepicker-inline`, disabling dates "asynchronously" (e.g., after the component has been created) breaks firing of `bsValueChange`.

**Plunker/StackBlitz that reproduces the issue**:

https://stackblitz.com/github/evilstiefel/ngx-bootstrap-datepicker-repro

**Versions of ngx-bootstrap, Angular, and Bootstrap**:

ngx-bootstrap: 7.0.0

Angular: 12.2.0

Bootstrap: 5.1.0

**Expected behavior**

I would expect to be able to set the `datesDisabled` property at any time, disabling dates, and not breaking change detection or the EventEmitter for `bsValueChange`.

SvetlanaMuravlova:
fix(datepicker): fixed inline datepicker subscriptions
close #6263 

What to test:
In inline datepicker and daterangepicker was fixed subscriptions

Preconditions:
use local version from this branch for reproducing this fix from issue

How to test:
on local version try reproducing as in issue example


apotapcukv (dismissed):

src/datepicker/bs-datepicker-inline.component.ts
```diff
@@ -131,6 +133,13 @@ export class BsDatepickerInlineDirective implements OnInit, OnDestroy, OnChanges
     }
   }
 
+  unsubscribeSubscriptions() {
+    if (this._subs?.length) {
+      this._subs.map(sub => sub.unsubscribe());
+    }
+    this._subs = [];
```
valorkin:
```suggestion
```

src/datepicker/bs-datepicker-inline.component.ts
```diff
@@ -131,6 +133,13 @@ export class BsDatepickerInlineDirective implements OnInit, OnDestroy, OnChanges
     }
   }
 
+  unsubscribeSubscriptions() {
+    if (this._subs?.length) {
+      this._subs.map(sub => sub.unsubscribe());
```
valorkin, line 138:
```suggestion
      this._subs.map(sub => sub.unsubscribe());
      this._subs.length = 0;
```

valorkin (changes requested):
unsubscribe should be in all versions I suppose

valorkin (approved):

Issue #5705, opened 2020-03-05
bsDatepicker adaptivePosition does not respect x-axis and overflows.


Totati:
**Bug description:**
Datepickers on the side of the page are still centered relative to the input, this causes the pickers to overflow the body. On the right it causes scrolling, on the left if just unusable.

[StackBlitz ](https://stackblitz.com/edit/ngx-bootstrap-datepicker-adaptive-position)that reproduces the issue.

**Versions of ngx-bootstrap, Angular, and Bootstrap:**
ngx-bootstrap: "ngx-bootstrap": "^5.5.0"
Angular: "@angular/core": "^9.0.5",
Bootstrap: "bootstrap": "^4.4.1",
Build system: Angular CLI, System.js, webpack, starter seed: 
Angular CLI: 9.0.5
Node: 10.16.0
OS: win32 x64

**Expected behavior**
Adaptive positioning should respect the window bounding box too, and position the datepicker correctly on the left/right if there's enought space.

nightskyre:
I am running into this same issue.

sharathdaniel:
Any solution for this?

casper5822:
Same issue using angular 12 with bootstrap 5.  Overflow-x of the body is not respected.

"bootstrap": "^5.1.3",
"ngx-bootstrap": "^7.1.2"

PR #6408, opened 2022-01-12
feat(datepicker): prevent overflow if adaptivePosition is set to true

niklaushug:
## Motivation
If the calendar of `[datepicker]` or `[daterangepicker]` is wider than `<input>` and if <input> is positioned on the very left or right of the viewport, it is likely that the calendar is not completely accessible.

## Solution 
By using modifier `preventOverflow` of `PositioningService` in `bs-datepicker-container.component.ts` and `bs-daterangepicker-container.component.ts` an overflow is prevented. I used the same logic like it was used for `popover.directive.ts` and `tooltip.directive.ts`.

## PR Checklist
 - ✓ read and followed the [CONTRIBUTING.md](https://github.com/valor-software/ngx-bootstrap/blob/development/CONTRIBUTING.md) guide.
 - ✓ built and tested the changes locally. --> Manual test on local server and automatic test by executing `npm run test`.
 - X added/updated tests.
 - ✓ added/updated API documentation --> see bs-datepicker.config.ts
 - X added/updated demos --> A demo for [adaptive-position](http://localhost:4200/ngx-bootstrap/#/components/datepicker?tab=overview#adaptive-position) exist already .

## Related issue
- https://github.com/valor-software/ngx-bootstrap/issues/5705

### DIFF
--- a/src/datepicker/bs-datepicker.config.ts
+++ b/src/datepicker/bs-datepicker.config.ts
@@ -183,4 +183,8 @@ export class BsDatepickerConfig implements DatepickerRenderOptions {
    * Set current hours, minutes, seconds and milliseconds for bsValue
    */
   initCurrentTime?: boolean;
+  /**
+   * Set allowed positions of container.
+   */
+  allowedPositions = ['top', 'bottom'];
 }
--- a/src/datepicker/themes/bs/bs-datepicker-container.component.ts
+++ b/src/datepicker/themes/bs/bs-datepicker-container.component.ts
@@ -69,8 +69,15 @@ export class BsDatepickerContainerComponent extends BsDatepickerAbstractComponen
 
   ngOnInit(): void {
     this._positionService.setOptions({
-      modifiers: { flip: { enabled: this._config.adaptivePosition } },
-      allowedPositions: ['top', 'bottom']
+      modifiers: {
+        flip: {
+          enabled: this._config.adaptivePosition
+        },
+        preventOverflow: {
+          enabled: this._config.adaptivePosition
+        }
+      },
+      allowedPositions: this._config.allowedPositions
     });
 
     this._positionService.event$?.pipe(take(1))
--- a/src/datepicker/themes/bs/bs-daterangepicker-container.component.ts
+++ b/src/datepicker/themes/bs/bs-daterangepicker-container.component.ts
@@ -77,8 +77,15 @@ export class BsDaterangepickerContainerComponent extends BsDatepickerAbstractCom
 
   ngOnInit(): void {
     this._positionService.setOptions({
-      modifiers: { flip: { enabled: this._config.adaptivePosition } },
-      allowedPositions: ['top', 'bottom']
+      modifiers: {
+        flip: {
+          enabled: this._config.adaptivePosition
+        },
+        preventOverflow: {
+          enabled: this._config.adaptivePosition
+        }
+      },
+      allowedPositions: this._config.allowedPositions
     });
 
     this._positionService.event$?.pipe(take(1))


valorkin (approved):
looks great to me. thanks for your help

talamaska:
I think this should be reverted. It is breaking a behavior from v6 and v7, and maybe even earlier. This was working fine. Now if your modal is smaller and the container had overflow auto, the date picker positions on top of the clicked field, totally hiding the text field and creating awkward experience. I don't think the overflow of the container should somehow break the position of the datepicker. It should look only at the position relative to the body, so that it avoids going offscreen, not trying to avoid going of an overflowed container

talamaska:
This is on v 8.0.0
image
image

talamaska:
This is even weirder on v 7.1.2
image
 Here it can go outside of the overflow modal, but tries to alternate bottom position with left and right, while it should alternate only top and bottom. But at least is not positioning itself over the field

niklaushug:
I am sorry that this change has bad implication. Have you checked already the configuration for `allowedPositions`?

Exposing more [PositioningOptions](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/positioning/models/index.ts#L29) to `BsDatepickerContainerComponent` seems to me a reasonable approach for flexibility. It is the same approach like it is used in [popover.directive.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/popover/popover.directive.ts#L158) and [tooltip.directive.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/tooltip/tooltip.directive.ts#L278).

Forcing options `flip` and `preventOverflow` to have the same value seems to me a good choice. May should option `boundariesElement` exposed as well?

Exposing more options is not the real cause. I reason if the problem could be located in `positioning.service.ts`?

talamaska:
Seems like allowedPositions was not documented on the web site for the v 8.
Also it doesn't seem to exist in 7.1.2 where I have things positioned on the side :)

On version 8.0.0 the preventOverflow is enabled if I choose adaptivePosition, which is not what I want. they should be controlled separately. Maybe if I have those exposed and boundariesElement, I might have the desired behavior, because I would be able to use adaptivePositon and set the boundariesElement  to be the window not the scroll element or some other container. In addition the prevent overflow also should be a separate config. The current v8 datePicker doesn't seem to specify a default boundariesElement, I wonder how it is passed and what uses the positioning service.

niklaushug:
If `boundariesElement` is not set the default in [preventOverflow.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/positioning/modifiers/preventOverflow.ts#L23)  is used.

I am open for decoupling options `flip.enabled` and `preventOverflow.enabled`. Option `adaptivePosition` could still be used as fallback and for compatibility.

Option `adaptivePosition` was introduced by @Domainv . May he wants to share his thoughts?

It would make sense to handle [popover.directive.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/popover/popover.directive.ts#L158) and [tooltip.directive.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/tooltip/tooltip.directive.ts#L278) the same way.

talamaska:
> If boundariesElement is not set the default in [preventOverflow.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/positioning/modifiers/preventOverflow.ts#L23) is used.

Maybe That's the issue with the date picker then.

Wondering why do you invent your logic for positioning and not using the Angular CDK overlay and portal stuff?

valorkin:
> > If boundariesElement is not set the default in [preventOverflow.ts](https://github.com/valor-software/ngx-bootstrap/blob/3e818fdffb77ab3d88e9ea1b27d448bed9437343/src/positioning/modifiers/preventOverflow.ts#L23) is used.
> 
> Maybe That's the issue with the date picker then.
> 
> Wondering why do you invent your logic for positioning and not using the Angular CDK overlay and portal stuff?
> 

Oh, that's an easy question, our positioning is oldef than cdk

talamaska:
Ok, so should I expect a change that will help me resolve my issues with the datepicker positioning.

valorkin:
for v8 started a serious of refactoring activities, what kind of change you would love to see?

talamaska:
at least the exposing of those props we discussed above -  boundariesElement and make preventOverflow configurable, by default it could be linked with the adaptivePosition, but I need to specify the boundaries element of the datepicker, to avoid limiting it's position to a scroll element, I would rather limit it in a window.

valorkin:
let us take a look

talamaska:
While looking into this, is there a way to have an adaptive position of the dropdown? I know about the dropup prop, but i think I can't set it just before the dropdown opens.

talamaska:
@valorkin @niklaushug Is there an issue in order to track the progress? It is really important for me to know the progress and the roadmap for 8.x

PR #6446, opened 2022-03-19
feat(button): allow any type for the value binding of the btnRadio directive

### LINKED ISSUE #6442: btnRadio should be declared as type "any"
In radio button groups, the btnRadio attribute is able to accept values of any type - objects, enums, numbers, booleans all work fine at runtime. However, my IDE (WebStorm) constantly reminds me that `btnRadio` is declared as type `string` in `button-radio.directive.d.ts`, so my the following syntax is not strictly valid, even though it works:
```
<div class="btn-group" btnRadioGroup [(ngModel)]="radioModel">
    <label class="btn btn-success mb-0" [btnRadio]="true">Yes</label>
    <label class="btn btn-success mb-0" [btnRadio]="false">No</label>
  </div>
```
This has come up before and it is acknowledged that non-string types are compatible with `btnRadio`: https://github.com/valor-software/ngx-bootstrap/issues/1808.

Any chance we could get `btnRadio` declared as type `any` in `button-radio.directive.d.ts` to reflect it's true behaviour?

wwarby:
There seems to be no reason that the value property on the button radio directive must be a string. Allowing binding to boolean, object or other types for this property has been acknowledged as valid in previous issues in this repo, but since the property is declared as type `string`, some IDEs such as WebStorm will recognise the type discrepancy and report a coding error if you attempt to bind a non-string to this property. The proposed changes in this pull request change the property type to `any` so that all types can be used when binding to the value.

closes #6442


williamxsp (dismissed):

valorkin:
I would try `unknown` type instead of `any`

valorkin (approved):
awesome, thanks!

PR #6451, opened 2022-04-05
fix: add bs-datepicker.css file to package.json exports

### LINKED ISSUE #6449: Yarn PnP cannot import bs-datepicker.css
On a working Angular codebase which has a less import statement `@import (less) 'ngx-bootstrap/datepicker/bs-datepicker.css';`, if I switch to using Yarn PNP, webpack loader throws an error that it cannot find the file.

If I edit the generated "package.json"  of ngx-boostrap and  add `"./datepicker/bs-datepicker.css": "./datepicker/bs-datepicker.css"`  to the "exports" property, the error goes away and seems to be resolved.

This is probably not a bug in Yarn itself as the Node docs (and the original proposal) seem to suggest that the "exports" field is used to conceal internals and until Yarn PNP (using the conventional 'node_modules" folder) there was just no simple way to enforce it.:
https://nodejs.org/api/packages.html#packages_exports
https://github.com/jkrems/proposal-pkg-exports/

salmoro:
This PR enables importing "bs-datepicker.css" file when using Yarn with PnP. When declared, the "exports" field in a package.json is meant to conceal and exclude all unspecified files or directories. Until now the exclusion of "bs-datepicker.css" wasn't an apparent issue because no other package manager/strategy is able to enforce disallowing importing files not specified in the "exports" field but Yarn with PnP enforces it.

https://nodejs.org/api/packages.html#packages_exports
https://github.com/jkrems/proposal-pkg-exports/
https://angular.io/guide/creating-libraries#managing-assets-in-a-library

fixes #6449

### DIFF
--- a/src/root/package.json
+++ b/src/root/package.json
@@ -80,6 +80,9 @@
       "node": "./datepicker/fesm2015/ngx-bootstrap-datepicker.mjs",
       "default": "./datepicker/fesm2020/ngx-bootstrap-datepicker.mjs"
     },
+    "./datepicker/bs-datepicker.css": {
+      "style": "./datepicker/bs-datepicker.css"
+    },
     "./dropdown": {
       "types": "./dropdown/ngx-bootstrap-dropdown.d.ts",
       "esm2020": "./dropdown/esm2020/ngx-bootstrap-dropdown.mjs",


valorkin (approved):

valorkin:
thanks for your help:)

PR #6457, opened 2022-04-25
fix(datepicker): fixed timepicker updates in datepicker with timepicker

### LINKED ISSUE #6433: DatePicker withTimePicker
When using the _Datepicker_ with _withTimePicker_ option the only way to set the time is when the initial input value is empty.  You must then set the time before selecting the date.  It is not possible to reset the time after re-opening the date picker, even if the time is changed and a different day is selected the time value will not change.
The problem can be found using the demo on the ngx-bootstrap site (v8.0.0):
[Date-picker-with-time-picker](https://valor-software.com/ngx-bootstrap/#/components/datepicker?tab=overview#with-timepicker)

SvetlanaMuravlova:
fix(datepicker): fixed timepicker updates in datepicker with timepicker
closes #6433


wedgberto:
I checked the demo site and can confirm that the daterangepicker does now emit time changes, however the picker has become painful to use because it is not possible to select a complete datetime range without the picker closing down multiple times!
daterangepicker-time-dance

SvetlanaMuravlova:
> I checked the demo site and can confirm that the daterangepicker does now emit time changes, however the picker has become painful to use because it is not possible to select a complete datetime range without the picker closing down multiple times! daterangepicker-time-dance

I checked your issue, and i can confirm it. But as i can see it is default behavior of datepicker and daterangepicker, so you can open a new issue as enhancement for it  and we will discuss it.

Tea-Ice:
@valorkin Any chance of merging this in?

queejie:
It seems like the bug that I'm experiencing is identified here, but the fix was not incorporated?  You can't scroll the time without the dialog closing, and if you pick a date before setting the time, it closes before you can do anything else.

SvetlanaMuravlova:
> It seems like the bug that I'm experiencing is identified here, but the fix was not incorporated? You can't scroll the time without the dialog closing, and if you pick a date before setting the time, it closes before you can do anything else.

yeah, it is, but it could be resolved in another pr, cause it is the native behaviour of element

SvetlanaMuravlova:
it was resolved in previous pr and as result it is fixed since version 9.0.0

Issue #6195, opened 2021-07-05
Modal is not always hidden after calling .hide()


gregoire-rouillon:
**Bug description**

Calling the `.hide()` method on a modal will not hide it if the opening animation (150ms by default) is not finished yet.

**Plunker/StackBlitz that reproduces the issue**

StackBlitz: https://stackblitz.com/edit/ngx-bootstrap-csh27l?file=src%2Fapp%2Fapp.component.ts

**Versions of ngx-bootstrap, Angular, and Bootstrap**

ngx-bootstrap: 6.2.0
Angular: 10.0.2
Bootstrap: 3.3.7

**Expected behavior**

The modal should hide

dezsiszabi:
I see there is a merged PR for this, so I assume it'll be available in 10.x.
Any chance of backporting this to 9.x and 8.x?
I encountered the same issue in my project (which is still on 8.x) and having this fix would be really nice.

renaldas-kerpe-arria:
Is this fixed in 9.x or 10.x?

renaldas-kerpe-arria:
Which version was this fixed in? I see the same issue in 8.x.

Looks like this issue might have been re-introduced https://github.com/valor-software/ngx-bootstrap/issues/6195

gregoire-rouillon:
Fixed by #6518 in 10.1.0

Issue #5840, opened 2020-06-29
bsModalRef sets an unneeded timeout on hide()


vvolodin:
Bug description:
Closing a dialog via `bsModalRef`'s `hide()` is setting a timeout for TRANSITION_DURATIONS.MODAL before actually doing anything to start hiding the dialog causing an unnecessary delay. Total time to close the dialog is 750 ms (TRANSITION_DURATIONS.MODAL*2 + TRANSITION_DURATIONS.BACKDROP), at least 300 of these happening before performing any animations. This is different when clicking on a backdrop, there's no additional delay there.

Plunker/StackBlitz that reproduces the issue:
https://valor-software.com/ngx-bootstrap/#/modals
Can be seen on ngx-bootstrap documentation website. Closing a dialog via "X" button is causing a 300 ms delay before starting animations (vs 0 delay when clicking on a backdrop).

Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: 5.6.2

Angular: 9.1.3

Bootstrap: 4.3.1

Build system: Angular CLI
Do not cause additional delays before starting close animations for a modal.

Issue #4270, opened 2018-04-26
modal.hide() - isShown=false needs to be the last thing happening


ctrl-brk:
When modal.hide() is called the isShown property gets set immediately as you can see here https://github.com/valor-software/ngx-bootstrap/blob/49ee88ed35349d980e53b993e0cfc4e53e4d206d/src/modal/modal-container.component.ts#L127 but all the other elements are still visible at this time. So if you base your logic (ex: ngIf visibility) on that property then the visual appearance is not pleasant. First your element disappears from the modal and only then the modal itself hides.

PR #6518, opened 2022-10-24
fix(modal): hide works even if show hasn't finished

### LINKED ISSUE #6439: Modal created via BsModalService is not hidden after calling .hide() due to animation delay
Bug description:

If you call hide on a modal created via BsModalService.show(Component) before the animation has completed, the modal doesn't hide and is left on screen. 

It looks like a similar issue was raised back in [2018](https://github.com/valor-software/ngx-bootstrap/issues/3711) and a fix implemented. I have been able to do a similar thing to the fix and put the hide inside a setTimeout with a delay equivalent to TRANSITION_DURATIONS.MODAL (300ms)

ngx-bootstrap: 8.0.0

Angular: 13.1.4

Bootstrap: 3.3.7

Expected behavior:

Calling hide hides the modal regardless of animation state

alejandrocoding:
This PR fix the problem we face when calling `hide()` on a modal reference before the modal animation has finished, in which case `hide()` has no effect into the displayed modal, staying open.

This change will enable us to call `show()` and right after `hide()` - and it will work as expected. Unit test is added.

The current workaround is adding `setTimeout` to call `hide()` in those cases that we might want to hide before loading the modal content is finished (cache data, quick api response, e2e tests...). This may impact the codebase and e2e tests, adding unnecessary timeouts and async awaits.

There is also another way of hiding the modal in those short-time shows, which is calling the `hide` method from the service passing the modal reference id, but still feels like the hide method from the modal reference should act the same way. These two ways and the issue are shown in the demo link below:

Issue demo: https://stackblitz.com/edit/ngx-bootstrap-modal-show-hide-issue-demo

before:
```
const ref = this.modalService.show(MyModalComponent);
// ref.hide(); // it wont work...
setTimeout(() => ref.hide(), TRANSITION_DURATIONS.MODAL);
```

after:
```
const ref = this.modalService.show(MyModalComponent);
ref.hide(); // it works!
```

Fixes https://github.com/valor-software/ngx-bootstrap/issues/6439 https://github.com/valor-software/ngx-bootstrap/issues/6195 https://github.com/valor-software/ngx-bootstrap/issues/5840 https://github.com/valor-software/ngx-bootstrap/issues/4270
______
_I took the liberty of improving/fixing the modal.service.spec. It seems that the done callbacks are not needed. Last test was not working, it was not entering the subscribe callback, so not executing the expect, making the test pass without validating the test case._

# PR Checklist

 - [X] read and followed the [CONTRIBUTING.md](https://github.com/valor-software/ngx-bootstrap/blob/development/CONTRIBUTING.md) guide.
 - [X] added/updated tests.
 - [ ] added/updated demos.

### DIFF
--- a/src/modal/bs-modal.service.ts
+++ b/src/modal/bs-modal.service.ts
@@ -134,7 +134,7 @@ export class BsModalService {
       .provide({ provide: BsModalRef, useValue: bsModalRef })
       .attach(ModalContainerComponent)
       .to('body');
-    bsModalRef.hide = () => modalContainerRef.instance?.hide();
+    bsModalRef.hide = () => this.hide(bsModalRef.id);
     bsModalRef.setClass = (newClass: string) => {
       if (modalContainerRef.instance) {
         modalContainerRef.instance.config.class = newClass;
--- a/src/modal/modal-container.component.ts
+++ b/src/modal/modal-container.component.ts
@@ -135,7 +135,7 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
   }
 
   hide(): void {
-    if (this.isModalHiding || !this.isShown) {
+    if (this.isModalHiding) {
       return;
     }
 
--- a/src/modal/testing/modal.service.spec.ts
+++ b/src/modal/testing/modal.service.spec.ts
@@ -7,17 +7,17 @@ import { BsModalService, ModalModule } from '../index';
 @Component({ template: '<div>Dummy Component</div>' })
 class DummyComponent {
   // eslint-disable-next-line @typescript-eslint/no-empty-function,@typescript-eslint/no-unused-vars
-    constructor(modalService: BsModalService) { }
+  constructor(modalService: BsModalService) { }
 }
 
 @Component({
-    template: '<div>Test Component</div>'
+  template: '<div>Test Component</div>'
 })
-class TestModalComponent { }
+export class TestModalComponent { }
 
 @NgModule({
-    declarations: [TestModalComponent],
-    entryComponents: [TestModalComponent]
+  declarations: [TestModalComponent],
+  entryComponents: [TestModalComponent]
 })
 export class TestModule { }
 
@@ -37,21 +37,19 @@ describe('Modal service', () => {
     fixture.detectChanges();
   });
 
-  it('should return random id on spin up a new modal', done => {
+  it('should return random id on spin up a new modal', () => {
     modalService.onShown.subscribe((data) => {
       expect(data.id).toBeTruthy();
-      done();
     });
 
     modalService.show(TestModalComponent);
   });
 
-  it('should return different random id when open nested modal without specifying id', done => {
+  it('should return different random id when open nested modal without specifying id', () => {
     modalService.onShown.pipe(
       pairwise(),
       tap(([firstData, secondData]) => {
         expect(firstData.id).not.toBe(secondData.id);
-        done();
       })
     ).subscribe();
 
@@ -59,12 +57,11 @@ describe('Modal service', () => {
     modalService.show(TestModalComponent);
   });
 
-  it('should return id in config when specified', done => {
+  it('should return id in config when specified', () => {
     const id = 20;
 
     modalService.onShown.subscribe((data) => {
       expect(data.id).toBe(id);
-      done();
     });
 
     modalService.show(TestModalComponent, { id });
@@ -72,6 +69,7 @@ describe('Modal service', () => {
 
   it('should return id when hide modal', () => {
     const id = 20;
+    jest.useFakeTimers();
 
     modalService.onHidden.subscribe((data) => {
       expect(data.id).toBe(id);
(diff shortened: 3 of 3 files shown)


src/modal/bs-modal.service.ts
```diff
@@ -134,7 +134,7 @@ export class BsModalService {
       .provide({ provide: BsModalRef, useValue: bsModalRef })
       .attach(ModalContainerComponent)
       .to('body');
-    bsModalRef.hide = () => modalContainerRef.instance?.hide();
+    bsModalRef.hide = () => this.hide(bsModalRef.id);
```
alejandrocoding, line 137: This is how the modal is hide when calling from the service itself instead of from the reference. The action is immediate.

src/modal/modal-container.component.ts
```diff
@@ -135,7 +135,7 @@ export class ModalContainerComponent implements OnInit, OnDestroy {
   }
 
   hide(): void {
-    if (this.isModalHiding || !this.isShown) {
+    if (this.isModalHiding) {
```
alejandrocoding, line 138: It should hide modal even if it is not shown, as shown is set after transition delay. We could keep that if we refactor the way this is set.

src/modal/testing/modal.service.spec.ts
```diff
@@ -7,17 +7,17 @@ import { BsModalService, ModalModule } from '../index';
...
-    constructor(modalService: BsModalService) { }
+  constructor(modalService: BsModalService) { }
 }
 
 @Component({
-    template: '<div>Test Component</div>'
+  template: '<div>Test Component</div>'
 })
-class TestModalComponent { }
+export class TestModalComponent { }
```
alejandrocoding, line 16: This should not be exported out of the test but it seems angular service call out an error in the below TestModule declarations array without the export. I can remove it if this is a concern.

src/modal/testing/modal.service.spec.ts
```diff
@@ -37,47 +37,69 @@ describe('Modal service', () => {
...
       expect(data.id).toBe(id);
-      done();
     });
 
     modalService.show(TestModalComponent, { id });
   });
 
   it('should return id when hide modal', () => {
     const id = 20;
+    jest.useFakeTimers();
```
alejandrocoding, line 75: Without waiting for the timers, onHidden is not triggered and the expect inside does not execute.

alejandrocoding:
All tests are passing in local:

src/modal/testing/modal.service.spec.ts
```diff
@@ -37,47 +37,69 @@ describe('Modal service', () => {
     fixture.detectChanges();
   });
 
-  it('should return random id on spin up a new modal', done => {
+  it('should return random id on spin up a new modal', () => {
     modalService.onShown.subscribe((data) => {
       expect(data.id).toBeTruthy();
-      done();
```
dianjuar: In my experience, your tests must call the `done` callback when you test async operations.

The problem is that a test without an `expect` will pass as successfully. For instance, this counts as a successful test:

```ts
it('should pass', () => {
  const foo = 'bar';
})
```

The test runs, and if it doesn't find any `expect` at the end, it will pass as valid.
This example will also pass, even though we want it to fail.

```ts
it('should fail', () => {
  setTimeout(() => {
    expect(true).toBe(false);
  }, 3000);
});
```

It will run, put an async operation on queue, finds the end of the test, no `expect` found, test successful, and our test never takes our assert into account.

---

To tell the test that the assert executed after an async operation, we must use the `done` callback

```ts
it('should fail', (done) => {
  setTimeout(() => {
    expect(true).toBe(false);
    done();
  }, 3000);
});
```

src/modal/testing/modal.service.spec.ts
```diff
@@ -37,47 +37,69 @@ describe('Modal service', () => {
...
+    jest.runAllTimers();
+  });
+
+  it('should hide modal even if invoked right after show', () => {
+    const id = 25;
+    const onHideSpy = jest.spyOn(modalService.onHide, 'emit');
+    const onHiddenSpy = jest.spyOn(modalService.onHidden, 'emit');
+    jest.useFakeTimers();
+
+    modalService.onHidden.subscribe((data) => {
```
dianjuar: I see that this test is taken by the test `should return id when hide modal`.

I would suggest skipping that test and focusing on testing one thing.
If it fails you would isolate what the problem is, otherwise the problem can be in several places.

I found [this](https://stackoverflow.com/questions/235025/why-should-unit-tests-test-only-one-thing/235032#235032) on the internet and explain better my point

src/modal/testing/modal.service.spec.ts
```diff
@@ -37,47 +37,69 @@ describe('Modal service', () => {
...
+
+    modalService.onHide.subscribe((data) => {
+      expect(data.id).toBe(id);
+    });
+
+    const bsRef = modalService.show(TestModalComponent, { id });
+    bsRef.hide();
+
+    jest.runAllTimers();
+    expect(onHideSpy).toHaveBeenCalled();
```
dianjuar: I would suggest to make two different test to test `onHide` and `onHidden`

dianjuar (commented):
I have a few recommendations for the unit test.

Awesome Work 🎉

src/modal/testing/modal.service.spec.ts
alejandrocoding: Thanks for clarification! Bringing them back!

src/modal/testing/modal.service.spec.ts
alejandrocoding: Got it, done.

src/modal/testing/modal.service.spec.ts
alejandrocoding: Thank you for the detailed message!

src/modal/testing/modal.service.spec.ts
alejandrocoding: done

dianjuar (approved):

valorkin:
looks great, thank you so much for your help)

src/modal/bs-modal.service.ts
renaldas-kerpe-arria: Sadly, this change broke the close interceptor 😥
https://github.com/valor-software/ngx-bootstrap/issues/6608

PR #6521, opened 2022-11-16
fix hide event registration leak

### LINKED ISSUE #6520: listenToTriggersV2 is leaking listeners
Consider the following code in listenToTriggersV2:

    // all listeners
    const listeners: Array<() => void> = [];
  
    // lazy listeners registration
    const _registerHide: Array<() => () => void> = [];
    const registerHide = () => {
      // add hide listeners to unregister array
      _registerHide.forEach((fn) => listeners.push(fn()));
      // register hide events only once
      _registerHide.length = 0;
    };
  
    // register open\close\toggle listeners
    parsedTriggers.forEach((trigger: Trigger) => {
      const useToggle = trigger.open === trigger.close;
      const showFn = useToggle ? options.toggle : options.show;
  
      if (!useToggle && trigger.close && options.hide) {
        const _hide = renderer.listen(target, trigger.close, options.hide);
        _registerHide.push(() => _hide);
      }
  
      if (showFn) {
        listeners.push(renderer.listen(target, trigger.open, () => showFn(registerHide)));
      }
    });
  
    return () => {
      listeners.forEach((unsubscribeFn: () => void) => unsubscribeFn());
    };

This line `const _hide = renderer.listen(target, trigger.close, options.hide);` registers a hide listener. 
At this moment listener `hide` already registered and needs to be unregistered.

Actual `registerHide` function might not be called at all because the show will never be triggered and `_hide` event subscriptions leaks.
I'm not really sure what this code should do but most probably you want to lazy-register hide. 
It means the code should be adjusted like this:

      if (!useToggle && trigger.close && options.hide) {
        _registerHide.push(() => renderer.listen(target, trigger.close, options.hide));
      }

ngx-bootstrap: 9.0.0
Angular: 14
Bootstrap: 5
Build system: Angular CLI

sherlock1982:
This make event registration truly lazy and fixes #6520


Alexveselun (approved):
Testing completed. No critical issues were found

valorkin (approved):
thank you

PR #6540, opened 2022-12-20
fix(datepicker): fixed disabled state on datepickers

### LINKED ISSUE #6471: Datepicker weird behavior when isDisabled is true.
Reproducible directly in the doc: https://valor-software.com/ngx-bootstrap/#/components/datepicker?tab=overview#disabled-datepicker

The datepicker, when enabled looks like this:
image

When isDisabled is set to true, the arrows to change month are hidden:
image

Furthermore, if we click on the year or month labels we are brought to a selection of year or month respectively, but the selection is also disabled, so we are stuck there.

This is not so bad in the datepicker, but as soon as you display a range that spans more than 1 month, the user will not be able to see where exactly the range ends:
image

I expected isDisabled to either not allow navigation at all by graying out the year/month labels also. But thinking about it, the range problem would still exist. So I suggest that isDisable prevents any new user input but keeps allowing all the calendar interactions.

Either way, I think that both options are needed and in any case the current state should be considered flawed.

### LINKED ISSUE #6487: Datepicker - Readonly attribute being set when isDisabled is set
Hello everyone, I spotted a weird behavior in the datepicker library :
It appears the readonly attribute's value is set to isDisabled's value when it should really be added or removed.

**Expected behavior**

Use setAttribute or removeAttribute respectively when isDisabled is set to true or false.

SvetlanaMuravlova:
close #6471 
close #6487


Alexveselun (approved):
Testing completed. No critical issues were found

Alexveselun (approved):

valorkin (approved):

Issue #4203, opened 2018-04-06
feat(datepicker): Support selection without implicit hide


sebek64:
### Bug description or feature request:

It'd be useful to allow disabling auto-hide on date selection. Use-case: a custom component combines datepicker with a custom timepicker. Visibility toggling is handled by the wrapping component. Currently, I need to call show() in onHidden (and asynchronously inside setTimeout, since it doesn't work directly).

### Versions of ngx-bootstrap, Angular, and Bootstrap:

ngx-bootstrap: 2.0.3

Angular: 4

Bootstrap: 3

mapsgeek:
thank you, you gave me a work around this, i need the datapicker to keep open by default and its value gets changed based on a click to another element which holds the range value, but the picker keeps closing and i have to click the input so it shows again.
it would be great if they can solve this so it keeps shown

neilime:
👍 it will be usefull to have a config option like hideOnSelect to disable auto-hide on date selection

jecarney:
Bumping this! When I use the workaround, the daterangepicker flickers on and off no matter how short the setTimeout.

gibahjoe:
Hi, what the update on this feature? Right now, I am using the datepicker with time and once the date is selected the picker closes even though a time has not been selected.

PR #6638, opened 2024-02-12
feat(datepicker): add apply and cancel buttons

keanan-meier:
# New feature description
This feature adds the option to enable an "Apply" and "Cancel" button to the datepicker. When enabled, the datepicker does not emit the change event from the view to the model until the value is applied by the new apply button. If the user chooses to cancel, the datepicker's value is reset back to the value of the model.

When this option is enabled it also keeps the datepicker open until the user applies, cancels, or dismisses the picker. This change improves the pickers behaviour when showTimePicker is also enabled as it allows the user to adjust both the date and time without the datepicker closing automatically after the date is updated. An issue discussed in #4203 .

The following new related config options have been added:

- showApplyButton
- showCancelButton
- applyButtonLabel
- cancelButtonLabel
- applyPosition

The css has also been slightly tweaked to improve appearance when all buttons are enabled. (Today, Clear, Apply and Cancel).

# PR Checklist

 - [X] added/updated tests.
 - [X] added/updated API documentation.
 - [X] added/updated demos.

### DIFF
--- a/apps/ngx-bootstrap-docs/src/ng-api-doc.ts
+++ b/apps/ngx-bootstrap-docs/src/ng-api-doc.ts
@@ -712,13 +712,14 @@ export const ngdoc: any = {
         name: 'showApplyButton',
         defaultValue: 'false',
         type: 'boolean',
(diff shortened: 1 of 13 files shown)


Maxie42:
I'd love to have this, would it be possible to fix the conflicts and get it merged? :)

PR #6743, opened 2025-05-14
fix: package.json types reference

joshvanallen:
Fixes the type references to enable apps that wants to use `moduleResolution:nodenext`

I have an [example repo](https://github.com/joshvanallen/ngx-bootstrap-module-resolution) that folks can use to play with that updates the package.json to the correct location. This uses 18.1.3 but the root package.json is still the same.

If we can, would love to apply this to 18.x.x as well.


joshvanallen:
@valorkin and @lexasq, brand new contributor to this repo so not too sure if you already get alerts for new PRs. Sorry if you do for the double alerts!

valorkin (approved):

valorkin:
thx for help!

PR #6810, opened 2026-05-21
feat(animations): replace @angular/animations with native CSS transitions for zoneless compatibility

### LINKED ISSUE #6784: Angular animations package is deprecated, ngx-bootstrap not working without
Since Angular 20.2, the @angular/animations package is [deprecated](https://angular.dev/guide/animations/migration).
Angular has a new set of [features for animations](https://angular.dev/guide/animations), mostly based on native CSS transitions and keyframes.

Removing the `BrowserAnimationsModule` or `provideAnimations()` from the providers results in errors for ngx-bootstrap components, such as dropdown and collapse.

As per the current documentation, Angular intents to remove the old animations package with Angular v23.

Migrating away from the animations package would allow developers to save on bundle size even today.
As of today, this issue prevents developers to migrate away.

### LINKED ISSUE #6808: Zoneless migration from version 21.2.0
Does this mean the angular application which still uses zone.js will not be supported going forward? As in if I need to upgrade to 21.2.0, I have to migrate my entire application to zoneless? And this path is empty as well:

https://github.com/valor-software/ngx-bootstrap/blob/v21.2.0/ZONELESS_MIGRATION_PLAN.md

DerekLiang:
## Summary
- Per Angular official doc https://angular.dev/api/animations/animation, `@angular/animations` is deprecated.
- Removes `@angular/animations` (and `BrowserAnimationsModule`) as a peer dependency entirely, eliminating the requirement to provide `BrowserAnimationsModule` in zoneless Angular applications
- Replaces all Angular animation triggers in `collapse`, `dropdown`, `datepicker`, and `typeahead` with native CSS `height` transitions driven by `requestAnimationFrame` + `transitionend` — the same pattern Angular's own migration guide recommends
- All `transitionend` handlers guard against bubbled child events (`e.target` + `e.propertyName` checks) and are properly cleaned up in `ngOnDestroy`
- Schematics updated to no longer inject `BrowserAnimationsModule` on `ng add`
- Fix animation issues （all animations are broken） that caused by zoneless migration

## Motivation

`@angular/animations` is deprecated as of Angular v20.2 and is incompatible with zoneless change detection — components using `AnimationBuilder` trigger zone re-entry via internal `setTimeout`/`requestAnimationFrame` calls, breaking zoneless apps. This was reported in issues #6784 and #6808.

## Changes by component

**Collapse** — `AnimationBuilder`-based player replaced with a `scrollHeight`-to-`0` CSS height transition; `transitionend` is guarded against child bubbling and `ngOnDestroy` cancels any pending rAF.

**Dropdown** — Both the inline (`BsDropdownDirective.playAnimation`) and dynamic-container (`BsDropdownContainerComponent`) paths use the same rAF-deferred transition; stale handlers cancelled on re-open and on destroy. `BsDropdownConfig` provider uses `Optional + SkipSelf` so nested dropdowns inherit parent config correctly.

**Datepicker / Daterangepicker** — `[@datepickerAnimation]` trigger removed from template; open animation implemented in `_animateDatepicker()` with `positionServiceEnable()` called in `transitionend` callback (replacing the `(@datepickerAnimation.done)` binding).

**Typeahead** — `[@typeaheadAnimation]` trigger removed; `_animateIn()` uses rAF deferral consistent with the other components; hardcoded `height: 100px` CSS override that was masking the dynamic height removed.

## Testing

All existing unit tests pass. Visual animations verified manually in the demo app for all four components.

Closes #6784
Closes #6808

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DerekLiang:
@lexasq Could you please provide feedback? Thank you!

lexasq:
Hi @DerekLiang, looks interesting, I'll check it out this week

lexasq:
I understand the idea, but we're just putting a lot more lifting on the ts than it could be. Would you consider to take a look at where we could utilize css native animations instead of rAF?

